PR #7, opened 2023-11-29
refacto(api): change the uri to respect the adr

tleon:
| Questions         | Answers
| Branch?           | dev
| Description?      | Fix the URI paths in order for them to respect the ADR
| Type?             | refacto
| Category?         | BO
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | CI 🟢 & UI tests 🟢
| UI Tests          | 
| Fixed issue or discussion?     | Fixes #34493 

### DIFF
--- a/src/ApiPlatform/Resources/ApiAccess.php
+++ b/src/ApiPlatform/Resources/ApiAccess.php
@@ -45,7 +45,7 @@ use PrestaShopBundle\ApiPlatform\Provider\QueryProvider;
 #[ApiResource(
     operations: [
         new Get(
-            uriTemplate: '/api-access/{apiAccessId}',
+            uriTemplate: '/apiAccess/{apiAccessId}',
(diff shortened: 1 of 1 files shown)


M0rgan01 (commented):
I find it strange to have camelcase in an API, in general it is rather advisable to have kebab-case in a URL. For example, Google better indexes URLs containing kebab case

> Consider using hyphens to separate words in your URLs, as it helps users and search engines identify concepts in the URL more easily. We recommend that you use hyphens (-) instead of underscores (_) in your URLs. 

https://developers.google.com/search/docs/crawling-indexing/url-structure

tleon:
https://github.com/PrestaShop/ADR/pull/35/files
The ADR describes it that way. Just following it

M0rgan01:
I know, but I just find it strange :)

PR #9, opened 2024-01-23
chore(ci): move tests and endpoint from core

tleon:
| Questions         | Answers
| Description?      | Move Api endpoints and associated tests into the module
| BC breaks?        | no
| Deprecations?     |no
| Fixed ticket?     | Fixes #35127
| How to test?      | CI 🟢


tests/Integration/ApiPlatform/DomainSerializerTest.php
```diff
@@ -0,0 +1,335 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Product\Command\AddProductCommand;
+use PrestaShop\PrestaShop\Core\Domain\Product\Query\GetProductForEditing;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductType;
+use PrestaShop\PrestaShop\Core\Domain\Shop\ValueObject\ShopConstraint;
+use PrestaShopBundle\ApiPlatform\DomainSerializer;
+use PrestaShopBundle\ApiPlatform\Resources\CustomerGroup;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+
+class DomainSerializerTest extends KernelTestCase
```
jolelievre, line 48: This part is a component from the core so it should remain in the core no need to copy it here, so you can remove this file

jolelievre (changes requested):

jolelievre (approved):
Thanks @tleon

PR #13, opened 2024-02-15
Update the resources with the custom operations, add the list hook endpoint

tleon:
| Questions         | Answers
| Description?      | Update the resources with the custom operations, add the list hook endpoint
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #34301
| How to test?      | CI 🟢

### DIFF
--- a/src/ApiPlatform/Resources/CartRule.php
+++ b/src/ApiPlatform/Resources/CartRule.php
@@ -29,16 +29,16 @@ declare(strict_types=1);
 namespace PrestaShop\Module\APIResources\ApiPlatform\Resources;
 
 use ApiPlatform\Metadata\ApiResource;
-use ApiPlatform\Metadata\Put;
 use PrestaShop\PrestaShop\Core\Domain\CartRule\Command\EditCartRuleCommand;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSPartialUpdate;
 use PrestaShopBundle\ApiPlatform\Processor\CommandProcessor;
 
 #[ApiResource(
     operations: [
-        new Put(
+        new CQRSPartialUpdate(
             uriTemplate: '/cart-rule',
             processor: CommandProcessor::class,
-            extraProperties: ['CQRSCommand' => EditCartRuleCommand::class]
+            CQRSCommand: EditCartRuleCommand::class
         ),
     ],
 )]
--- a/src/ApiPlatform/Resources/CartRule.php
+++ b/src/ApiPlatform/Resources/CartRule.php
@@ -30,14 +30,12 @@ namespace PrestaShop\Module\APIResources\ApiPlatform\Resources;
 
 use ApiPlatform\Metadata\ApiResource;
 use PrestaShop\PrestaShop\Core\Domain\CartRule\Command\EditCartRuleCommand;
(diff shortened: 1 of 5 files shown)


src/ApiPlatform/Resources/ApiAccess.php
```diff
@@ -68,13 +68,10 @@
                 ],
             ],
             provider: QueryProvider::class,
-            extraProperties: [
-                'query' => GetApiAccessForEditing::class,
-                'CQRSQuery' => GetApiAccessForEditing::class,
-                'scopes' => ['api_access_read'],
-            ]
+            CQRSQuery: GetApiAccessForEditing::class,
```
jolelievre, line 71: You can remove the `provider: QueryProvider::class,` when you use the `CQRSGet` it is already specified by default SAme suggestions for CQRS commands the processor is defined by default

tests/Integration/ApiPlatform/ListHooksTest.php
```diff
@@ -0,0 +1,83 @@
...
+
+    public static function tearDownAfterClass(): void
+    {
+        parent::tearDownAfterClass();
+        DatabaseDump::restoreTables(['hook']);
+    }
+
+    public function testListHooks(): void
+    {
+        $hooks = $this->generateHooks();
```
jolelievre, line 50: Since you generate the hooks and therefore can anticipate the content of the list I think it could be interesting to look through the JSON result and check the content (maybe you'll need an extra sortBy)

Also, could you check the other elements from the JSON, like total numbers (and orderBy if present) And the super stuff would be to add one example of filtering as well

src/ApiPlatform/Resources/Hook.php
```diff
@@ -67,28 +69,30 @@
...
             exceptionToStatus: [HookNotFoundException::class => 404],
             provider: QueryProvider::class,
-            extraProperties: [
-                'CQRSQuery' => GetHook::class,
-                'scopes' => ['hook_read'],
-            ]
+            CQRSQuery: GetHook::class,
+            scopes: ['hook_read']
+        ),
+        // PR module + inte module + convertir en custom ce qu'il reste en natif
```
matks, line 89: 🇫🇷

tleon: Oupsie !

src/ApiPlatform/Resources/CartRule.php
```diff
@@ -29,16 +29,14 @@
...
 use ApiPlatform\Metadata\ApiResource;
-use ApiPlatform\Metadata\Put;
 use PrestaShop\PrestaShop\Core\Domain\CartRule\Command\EditCartRuleCommand;
-use PrestaShopBundle\ApiPlatform\Processor\CommandProcessor;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSPartialUpdate;
 
 #[ApiResource(
     operations: [
-        new Put(
+        new CQRSPartialUpdate(
```
jolelievre, line 37: If it as the `PUT` method then we should use `CQRSUpdate`
`CQRSPartialUpdate` is used for `PATH` method

jolelievre (approved):
Thanks @tleon

PR #16, opened 2024-02-20
Update product endpoint query mapping to match new path for context values

jolelievre:
| Questions         | Answers
| Description?      | Update product endpoint query mapping to match new path for context values
| Type?             | refactor
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | ~
| Sponsor company   | ~
| How to test?      | ~


kpodemski:
@jolelievre, could you please explain how it works? If we work with an API, don't we explicitly provide the shopId? Where does the Context fit in?

M0rgan01 (approved):

PR #20, opened 2024-03-21
Add module GET endpoint and bulk PUT for status

jolelievre:
| Questions         | Answers
| Description?      | Add module GET endpoint and bulk PUT for status, also add paginated list
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | ~
| Sponsor company   | ~
| How to test?      | ~

### DIFF
--- a/src/ApiPlatform/Resources/Module/BulkModules.php
+++ b/src/ApiPlatform/Resources/Module/BulkModules.php
@@ -1,0 +1,58 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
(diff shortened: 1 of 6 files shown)


src/ApiPlatform/Resources/Module/BulkModules.php
```diff
@@ -0,0 +1,58 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
```
matks, line 8: AFL not OSL

jolelievre: Right but it's wrong in the whole module actually, I'll do a dedicated PR after to fix it all

matks (approved):
Approved but a PR to fix licensing will be needed afterwards

PR #33, opened 2024-10-16
Endpoint reset module

jinpresta:
| Questions         | Answers
| Description?      | Reset module, related to https://github.com/PrestaShop/PrestaShop/pull/37169
| Type?             |  improvement 
| BC breaks?        |  
| Deprecations?     | 
| Fixed ticket?     | ~
| Sponsor company   | ~
| How to test?      | ~

### DIFF
--- a/tests/Integration/ApiPlatform/ModuleEndpointTest.php
+++ b/tests/Integration/ApiPlatform/ModuleEndpointTest.php
@@ -61,6 +61,11 @@ class ModuleEndpointTest extends ApiTestCase
             'PUT',
             '/module/status/{technicalName}',
         ];
+
+        yield 'reset module' => [
+            'PUT',
+            '/module/{technicalName}/reset',
+        ];
     }
 
     public function testListModules(): array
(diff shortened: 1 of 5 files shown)


ps-jarvis:
Hello @jinpresta!

tests/Integration/ApiPlatform/ModuleEndpointTest.php
```diff
@@ -220,6 +225,46 @@ public function testUpdateModuleStatus(array $module): void
...
+            'enabled' => true,
+            'installed' => true,
+        ];
+        $this->assertEquals($expectedModuleInfos, $decodedResponse);
+    }
+
+    public function testResetModule_NotInstalled(): void
+    {
+        $module = [
+            'technicalName' => 'ps_notthere',
```
jolelievre, line 259: Actually, here you are testing the reset on a module that is not present at all This should return a 404 as I commented in the CQRS PR You can update this test once the CQRS command has been updated with the proper `ModuleNotFoundException`

Then you will have to add a mapping in the Api resource class (I added another comment explaining that), when it's done you can update this test to check that a 404 is correctly triggered when accessing non existent modules

You can even rename this one into `testModuleNotFound` and check that all the future endpoints will have this 404 behaviour (including the existing GET endpoint)

tests/Integration/ApiPlatform/ModuleEndpointTest.php
```diff
@@ -220,6 +225,46 @@ public function testUpdateModuleStatus(array $module): void
...
+        ];
+        $bearerToken = $this->getBearerToken(['module_read', 'module_write']);
```
jolelievre, line 261: The use case you aimed at testing is also interesting though, checking that a module is disabled or uninstalled and try to reset it which will throw a dedicated exception (that you already implemented) because the action is not allowed It makes sense but to have the proper conditions you will need to change the status of the module first: disable it using the status API, and uninstall it using the uninstall API

But since the uninstall is not available yet you can only assert the behaviour for disabled modules for now, the uninstalled case can be added later

jolelievre (approved):
Thanks @jinpresta

PR #35, opened 2024-10-23
Endpoint Uninstall module

dylanDenizonPresta:
| Questions         | Answers
| Description?      | Endpoint Uninstall module
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | ~
| Sponsor company   | ~
| How to test?      | ~

### DIFF
--- a/tests/Integration/ApiPlatform/ModuleEndpointTest.php
+++ b/tests/Integration/ApiPlatform/ModuleEndpointTest.php
@@ -61,6 +61,16 @@ class ModuleEndpointTest extends ApiTestCase
             'PUT',
             '/module/status/{technicalName}',
         ];
+
+        yield 'uninstall module' => [
+            'PUT',
+            '/module/{technicalName}/uninstall',
+        ];
+
+        yield 'bulk uninstall' => [
+            'PUT',
+            '/modules/uninstall',
+        ];
     }
 
     public function testListModules(): array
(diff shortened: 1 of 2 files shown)


tests/Integration/ApiPlatform/ModuleEndpointTest.php
```diff
@@ -301,6 +311,43 @@ public function restResetModuleNotActive(array $module): void
         self::assertResponseStatusCodeSame(400);
     }
 
+    public function testUninstallModule()
+    {
+        $module = [
+            'technicalName' => 'bankwire',
+            'deleteFile' => false
+        ];
```
jolelievre, line 319: No need for this variable, you can hard-code the value in your request parameter

tests/Integration/ApiPlatform/ModuleEndpointTest.php
```diff
@@ -301,6 +311,43 @@ public function restResetModuleNotActive(array $module): void
...
+        $bearerToken = $this->getBearerToken(['module_write']);
+        static::createClient()->request('PUT', sprintf('/module/%s/uninstall', $module['technicalName']), [
+            'auth_bearer' => $bearerToken,
+            'json' => [
+                'deleteFile' => $module['deleteFile'],
+            ],
+        ]);
+
+        self::assertResponseStatusCodeSame(204);
+
```
jolelievre, line 331: You should use `getModuleInfos` to check the updated status of the module

tests/Integration/ApiPlatform/ModuleEndpointTest.php
```diff
@@ -301,6 +311,43 @@ public function restResetModuleNotActive(array $module): void
...
+        static::createClient()->request('PUT', sprintf('/modules/uninstall'), [
+            'auth_bearer' => $bearerToken,
+            'json' => [
+                'modules' => $modules,
+                'deleteFile' => true,
+            ],
+        ]);
+
+        self::assertResponseStatusCodeSame(204);
+    }
```
jolelievre, line 349: You should check that now the API returns a 404 error code when we fetch the module's infos, maybe it's worth adding a `checkModuleNotFound` method similar to `getModuleInfos` except it returns void and it checks for the 404 error code

This can be reused in other test methods

jolelievre (commented):
Thanks @dylanDenizonPresta

tests/Integration/ApiPlatform/ModuleEndpointTest.php
jolelievre, line 1: Since there is now a test that removes some module files we should integrate the ResourceResetter in the tear down method of this test, so that everything is restored at the end of the tests

jolelievre:
Closed in favor of https://github.com/PrestaShop/ps_apiresources/pull/37

PR #49, opened 2025-05-22
Add discount endpoint

tleon:
| Questions         | Answers
| Description?      | Add new GET & POST Discount endpoint 
| BC breaks?        | no
| Deprecations?     |  no
| Fixed ticket?     | Fixes #38647
| How to test?      | CI green

### DIFF
--- a/tests/Integration/ApiPlatform/DiscountEndpointTest.php
+++ b/tests/Integration/ApiPlatform/DiscountEndpointTest.php
@@ -1,0 +1,187 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+declare(strict_types=1);
+
+namespace PsApiResourcesTest\Integration\ApiPlatform;
+
+use PrestaShop\PrestaShop\Core\Domain\Discount\Command\AddDiscountCommand;
+use PrestaShop\PrestaShop\Core\Domain\Discount\Query\GetDiscountForEditing;
+use Tests\Resources\Resetter\LanguageResetter;
+
+class DiscountEndpointTest extends ApiTestCase
+{
+    public const CART_LEVEL = 'cart_level';
+    public const PRODUCT_LEVEL = 'product_level';
+    public const FREE_GIFT = 'free_gift';
+    public const FREE_SHIPPING = 'free_shipping';
+    public const ORDER_LEVEL = 'order_level';
+
+    public static function setUpBeforeClass(): void
+    {
+        parent::setUpBeforeClass();
+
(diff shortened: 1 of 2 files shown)


tests/Integration/ApiPlatform/DiscountEndpointTest.php
```diff
@@ -0,0 +1,180 @@
...
+use Tests\Resources\Resetter\DiscountResetter;
+use Tests\Resources\ResourceResetter;
+
+class DiscountEndpointTest extends ApiTestCase
+{
+    public static function setUpBeforeClass(): void
+    {
+        parent::setUpBeforeClass();
+        (new ResourceResetter())->backupTestModules();
+        DiscountResetter::resetProducts();
```
jolelievre, line 44: There is no dedicated `DiscountResetter` class, but these resetter classes are shortcuts to the `DatabaseDump` class anyway. For products and languages that are used regularly it makes sens to have specific one but in this case you can simply use the generic approach like:
```suggestion
        DatabaseDump::restoreTables(['cart_rule', 'cart_rule_lang']);
```
(with all additional tables related to this domain)

tests/Integration/ApiPlatform/DiscountEndpointTest.php
```diff
@@ -0,0 +1,180 @@
...
+use Tests\Resources\Resetter\LanguageResetter;
+use Tests\Resources\Resetter\DiscountResetter;
+use Tests\Resources\ResourceResetter;
+
+class DiscountEndpointTest extends ApiTestCase
+{
+    public static function setUpBeforeClass(): void
+    {
+        parent::setUpBeforeClass();
+        (new ResourceResetter())->backupTestModules();
```
jolelievre, line 43:
```suggestion
```
Not sure it's needed for this domain

tests/Integration/ApiPlatform/DiscountEndpointTest.php
```diff
@@ -0,0 +1,180 @@
...
+        self::createApiClient(['discount_write', 'discount_read']);
+    }
+
+    public static function tearDownAfterClass(): void
+    {
+        parent::tearDownAfterClass();
+        DiscountResetter::resetProducts();
+        LanguageResetter::resetLanguages();
+        // Reset modules folder that are removed with the FR language
+        (new ResourceResetter())->resetTestModules();
```
jolelievre, line 56:
```suggestion
```

tests/Integration/ApiPlatform/DiscountEndpointTest.php
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    /**
+     * @depends testAddDiscount
```
jolelievre, line 60: Good! you used the `@depends` so `testAddDiscount` will be executed before in any case, but to ease the readability of this test class can you move this one after `testAddDiscount` please?

tests/Integration/ApiPlatform/DiscountEndpointTest.php
```diff
@@ -0,0 +1,180 @@
...
+     * @dataProvider discountTypesDataProvider
+     *
+     * @param string $type
+     * @param array $names
+     * @return int
+     */
+    public function testAddDiscount(string $type, array $names): int
+    {
+        // skip test if class does not exist
+        if (!class_exists(AddDiscountCommand::class)) {
```
jolelievre, line 100: Ah, I was searching how to do this, perfect!

I'm just curious, there's no need to `return;` after this method is called?

tests/Integration/ApiPlatform/DiscountEndpointTest.php
```diff
@@ -0,0 +1,180 @@
...
+        $this->assertEquals(
+            [
+            ],
+            $decodedResponse
+        );
+
+        return $discountId;
+    }
+
+    public function discountTypesDataProvider(): array
```
jolelievre, line 127: 👍

tests/Integration/ApiPlatform/DiscountEndpointTest.php
jolelievre: Aaah it actually throws an exception under the hood, so no need for a return

jolelievre (approved):
Nice job @tleon

PR #62, opened 2025-08-20
Fix folder name

MattKelvin:
| Questions         | Answers
| Description?      | We are currently checking the parameters file in the prestashop which is located in tmp. If we want to override it, we have to use the folder in the module (`$localeParameters`)
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | /
| Sponsor company   | Mademoiselle bio
| How to test?      | Edit `tests/local-parameters/parameters.yml` and execute `composer setup-local-tests`, your file will be ignored :'(

### DIFF
--- a/tests/EnvironmentBuilder.php
+++ b/tests/EnvironmentBuilder.php
@@ -99,7 +99,7 @@ class EnvironmentBuilder
         $configFolder = $prestashopDir . '/app/config/';
         $localeParameters = $moduleDir . '/tests/local-parameters/';
         foreach (['parameters.php', 'parameters.yml'] as $parameterFile) {
-            if (!file_exists($configFolder . $parameterFile) || $arguments['update-local-parameters']) {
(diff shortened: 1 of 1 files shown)


Touxten:
Hello,

You need to rebase your PR

Touxten (changes requested):
Need rebase

ps-jarvis:
This pull request seems to contain new translation strings. I have summarized them below to ease up review:

- `Modules.Apiresources.Admin` <span title="New domain!">⚠️</span>
    - [ ] `PrestaShop API Resources`
    - [ ] `Includes the resources allowing using the API for the PrestaShop domain, all endpoints are based on CQRS commands/queries from the Core and we APIPlatform framework is used as a base.`

(Note: this is an automated message, but answering it will reach a real human)

PR #81, opened 2025-09-30
Feature endpoints

Codencode:
| Questions         | Answers
| Description?      | 
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| Sponsor company   | @Codencode 
| How to test?      |


src/ApiPlatform/Resources/Feature/BulkFeatures.php
```diff
@@ -0,0 +1,54 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Feature\Command\BulkDeleteFeatureCommand;
+use PrestaShop\PrestaShop\Core\Domain\Feature\Exception\FeatureNotFoundException;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSUpdate;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Validator\Constraints as Assert;
+
+#[ApiResource(
+    operations: [
+        new CQRSUpdate(
+            uriTemplate: '/features/features/delete',
```
boherm, line 34: Its look weird no?
Maybe we should statuate with bulk actions to unified for all entities 🤔 

And with the same approach that Google Drive API do with a `batch` endpoint:

```suggestion
        new CQRSDelete(
            uriTemplate: '/features/batch',
```

src/ApiPlatform/Resources/Feature/Feature.php
```diff
@@ -0,0 +1,116 @@
...
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSGet;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSPartialUpdate;
+use PrestaShopBundle\ApiPlatform\Metadata\LocalizedValue;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Validator\Constraints as Assert;
+
+#[ApiResource(
+    operations: [
+        new CQRSGet(
+            uriTemplate: '/features/feature/{featureId}',
```
boherm, line 44: `/features/feature/...` looks weird also here, no?

```suggestion
            uriTemplate: '/features/{featureId}',
```

boherm (changes requested):

src/ApiPlatform/Resources/Feature/BulkFeatures.php
Codencode: I based my work on this:
https://github.com/PrestaShop/ps_apiresources/blob/20fb2231e1d92a0736aaf8d68fd28136810fa251/src/ApiPlatform/Resources/Attribute/BulkAttributeGroups.php#L33-L37

But your observation is absolutely right!
I'll make the change.
Thanks!

kpodemski: cc @jolelievre 

I think we need to add a few rules to ADR 🤔

src/ApiPlatform/Resources/Feature/Feature.php
Codencode: @boherm @kpodemski 
I made the changes

src/ApiPlatform/Resources/Feature/BulkFeatures.php
Codencode: @boherm

By using CQRSDelete instead of CQRSUpdate for the batch delete, I had to create the deleteBatch() method to be able to pass the $data parameter and execute the Request::METHOD_DELETE call.

Perhaps we could add this method directly here: https://github.com/PrestaShop/ps_apiresources/blob/dev/tests/Integration/ApiPlatform/ApiTestCase.php?

What do you think?
Maybe we could use a more appropriate name.

paulnoelcholot:
@Codencode no problem :)

I retry but I have a 400 Bad Request :/

image

PR #125, opened 2025-12-03
feat(TaxRulesGroup): Add partial update on status

Kaikina:
| Questions         | Answers
| Description?      | Add partial update for tax rules group on status property (enabled)
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | NA
| Sponsor company   | Evolutive
| How to test?      | Try via the swagger doc to edit the status of an existing tax rule group. Switch it to disabled then enabled.

### DIFF
--- a/src/ApiPlatform/Resources/TaxRulesGroup/TaxRulesGroup.php
+++ b/src/ApiPlatform/Resources/TaxRulesGroup/TaxRulesGroup.php
@@ -27,6 +27,7 @@ use ApiPlatform\Metadata\ApiResource;
 use PrestaShop\PrestaShop\Core\Domain\TaxRulesGroup\Command\AddTaxRulesGroupCommand;
 use PrestaShop\PrestaShop\Core\Domain\TaxRulesGroup\Command\DeleteTaxRulesGroupCommand;
 use PrestaShop\PrestaShop\Core\Domain\TaxRulesGroup\Command\EditTaxRulesGroupCommand;
+use PrestaShop\PrestaShop\Core\Domain\TaxRulesGroup\Command\SetTaxRulesGroupStatusCommand;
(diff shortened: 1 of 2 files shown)


kpodemski:
Hello @Kaikina 

As visible in the CI, the name you chose for one of the endpoints is invalid, it didn’t follow requirements from ADR. Could you make sure to fix it?

kpodemski (dismissed):
Blocking since CI is red, details in the message  above.

src/ApiPlatform/Resources/TaxRulesGroup/TaxRulesGroup.php
```diff
@@ -70,6 +71,18 @@
             CQRSQueryMapping: self::QUERY_MAPPING,
             scopes: ['tax_rules_group_write'],
         ),
+        new CQRSPartialUpdate(
+            uriTemplate: '/tax-rules-groups/{taxRulesGroupId}/update-status',
```
jolelievre: The Rector rule that checks the url format accepts a few action word
They are listed here https://github.com/PrestaShop/ps_apiresources/blob/534bdac75cb96a238cc3236027eecfd49107f465/tests/Rector/ApiResourceUriTemplateRector.php#L102

For new action the list can be extended, but we don't want to add too many different action words The idea is to try and have some consistent naming across all domains Same type of action = same wording

So before you add a new action you should check the available ones to see if one already matches your use case, and there is one actually:

```suggestion
            uriTemplate: '/tax-rules-groups/{taxRulesGroupId}/set-status',
```

Kaikina: Ok, thanks for the information !

boherm (approved):
Nice, LGTM. Thanks @Kaikina!

ps-jarvis:
PR merged, well done!

boherm:
Thanks!

PR #143, opened 2026-02-25
Add resource for CMS page

ga-devfront:
| Questions         | Answers
| Description?      | Add GetCmsPageForEditing CQRS endpoint
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| Sponsor company   | @PrestaShopCorp
| How to test?      | Ask a dev

### DIFF
--- a/src/ApiPlatform/Resources/CmsPage/CmsPage.php
+++ b/src/ApiPlatform/Resources/CmsPage/CmsPage.php
@@ -1,0 +1,88 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\CmsPage;
+
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ApiResource;
+use PrestaShop\PrestaShop\Core\Domain\CmsPage\Query\GetCmsPageForEditing;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSGet;
+use PrestaShopBundle\ApiPlatform\Metadata\LocalizedValue;
+
+#[ApiResource(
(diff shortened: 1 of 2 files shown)


ps-jarvis:
Hello @ga-devfront!

nicosomb:
@ga-devfront the CI is red, could you have a look please?

src/ApiPlatform/Resources/CmsPage/CmsPage.php
```diff
@@ -0,0 +1,88 @@
...
+
+    #[LocalizedValue]
+    public array $contents;
+
+    public bool $indexedForSearch;
+
+    public bool $displayed;
+
+    #[ApiProperty(openapiContext: ['type' => 'array', 'items' => ['type' => 'integer'], 'example' => [1, 3]])]
+    #[Assert\NotBlank(allowNull: true)]
```
mattgoud, line 70: `Assert` isn't imported (no `use Symfony\Component\Validator\Constraints as Assert;`), so this resolves to a non-existent attribute class and breaks on metadata reflection. it's also a no-op on a GET-only resource (validation only runs on write). please just remove this line.

tests/Integration/ApiPlatform/CmsPageEndpointTest.php
```diff
@@ -0,0 +1,74 @@
...
+
+    public function testGetCmsPage(): void
+    {
+        $cmsPage = $this->getItem('/cms-pages/1', ['cms_page_read']);
+
+        $this->assertArrayHasKey('cmsPageId', $cmsPage);
+        $this->assertEquals(1, $cmsPage['cmsPageId']);
+        $this->assertArrayHasKey('titles', $cmsPage);
+        $this->assertArrayHasKey('friendlyUrls', $cmsPage);
+        $this->assertArrayHasKey('enabled', $cmsPage);
```
mattgoud, line 71: the resource exposes `displayed`, not `enabled` — there's no `enabled` key in the response, so this assertion always fails. should be `$this->assertArrayHasKey('displayed', $cmsPage);`.

mattgoud (changes requested):
thanks @ga-devfront 🙏 — sheriff review. i ran an independent pass and also triggered the Claude pre-review (comment above); the two line up. the resource is nicely scoped and the `QUERY_MAPPING` is correct — i checked all 11 source fields against the `EditableCmsPage` getters in core and they match. a few blockers before merge:

**1. missing `exceptionToStatus` → 500 instead of 404**
the `#[ApiResource]` declares no `exceptionToStatus`, so `GET /cms-pages/{id}` on a non-existent page returns a 500. add:
```php
exceptionToStatus: [
    CmsPageNotFoundException::class => Response::HTTP_NOT_FOUND,
],
```
(`CmsPageNotFoundException` exists in core.) worth adding a 404 test alongside it.

**2. `#[Assert\NotBlank]` on `$shopIds` (l.70)** — there's no `use Symfony\Component\Validator\Constraints as Assert;` import, so this references an undefined attribute class; and on a GET-only resource a validation constraint does nothing anyway. simplest fix: remove the line. (inline)

**3. test asserts `enabled` (l.71)** but the resource field is `displayed` — there's no `enabled` key in the response, so the test fails. should be `displayed`. (inline)

minor: the identity entries in `QUERY_MAPPING` (`cmsPageId→cmsPageId`, `displayed→displayed`, …) are no-ops — only the renames (`localizedTitle→titles`, `shopAssociation→shopIds`, …) are needed; and the test could assert field values, not just key existence. the pre-review comment above has the full breakdown.

clean start otherwise — 1–3 and it's basically there. 👍

PR #149, opened 2026-03-11
Fix searchalias endpoints

### LINKED ISSUE #40526: Admin API: fix search alias endpoints
The endpoints for search alias have been put in experimental because they had issues
The problems come from missing feature from the core partially and also from the way the CQRS commands are built, so it may require multiple fixes on the core

But so far we coudln't implement REST endpoints (the create and update endpoints work, but we can't return the modified result)

boherm:
| Questions         | Answers
| Description?      | Fix search alias endpoints.<br/>I've added a normalizer to avoid a fix in the core.<br/>I don't think that we have this kind of behaviour for another endpoint.. it's specific as we use a string as identifier)
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/PrestaShop/issues/40526
| How to test?      | See https://github.com/PrestaShop/PrestaShop/issues/40526


mattgoud (dismissed):

src/ApiPlatform/Normalizer/GetSearchTermAliasesQuerySerializer.php
```diff
@@ -0,0 +1,62 @@
...
+    public function __construct(
+        private readonly RequestStack $requestStack,
+    ) {
+    }
+
+    public function denormalize(mixed $data, string $type, ?string $format = null, array $context = [])
+    {
+        $searchTerm = $data['searchTerm'];
+
+        $request = $this->requestStack->getCurrentRequest();
```
jolelievre, line 40: I think it deserves a comment to explain why we need to do this

src/ApiPlatform/Resources/SearchAlias/BulkDeleteSearchAliases.php
```diff
@@ -26,26 +26,46 @@
...
     operations: [
         new CQRSDelete(
-            uriTemplate: '/search-aliases/bulk-delete',
+            // Usually the identifier searchTerm would be in the URL, but since here it is a string this would
+            // conflict with the bulk url /search-aliases/bulk-delete (bulk-delete would be considered the identifier,
+            // vice versa)
+            // We will have only one endpoint for deletion, if there are multiple search terms to delete they will be
+            // passed in the JSON body, if there is only one search term to delete it will still need to be passed in
+            // the JSON body as an array with one element
+            uriTemplate: '/search-aliases',
```
jolelievre:
```suggestion
            uriTemplate: '/search-aliases/bulk-delete',
```
Since there is now only one endpoint relying on the DELETE method, I think we can keep the usual URI for bulk delete endpoint, as it cannot conflict with the signle DELETE endpoint anymore since it doesn't exist anymore

src/ApiPlatform/Resources/SearchAlias/BulkDeleteSearchAliases.php
```diff
@@ -26,26 +26,46 @@
...
 #[ApiResource(
     operations: [
         new CQRSDelete(
-            uriTemplate: '/search-aliases/bulk-delete',
+            // Usually the identifier searchTerm would be in the URL, but since here it is a string this would
+            // conflict with the bulk url /search-aliases/bulk-delete (bulk-delete would be considered the identifier,
+            // vice versa)
+            // We will have only one endpoint for deletion, if there are multiple search terms to delete they will be
+            // passed in the JSON body, if there is only one search term to delete it will still need to be passed in
+            // the JSON body as an array with one element
```
jolelievre, line 40:
```suggestion
            // the JSON body as an array with one element
            // As a consequence, there is no single deletion endpoint with the DeleteSearchTermAliasesCommand
```

src/ApiPlatform/Resources/SearchAlias/UpdateSearchAlias.php
```diff
@@ -86,6 +85,12 @@ class UpdateSearchAlias
     #[Assert\Length(min: 1, max: 255)]
     public ?string $newSearchTerm = null;
 
+    protected const QUERY_MAPPING = [
+        '[newSearchTerm]' => '[searchTerm]',
```
jolelievre: This one is not useful anymore thanks to the normalizer no?

jolelievre: Also, do you think it's possible to merge this one in the other ApiResource class?
Or there are some constraints that require keeping them separated?

src/ApiPlatform/Resources/SearchAlias/SearchAlias.php
```diff
@@ -42,16 +42,15 @@
             scopes: ['search_alias_read'],
             exceptionToStatus: [AliasNotFoundException::class => 404],
             CQRSQueryMapping: self::QUERY_MAPPING,
-            experimentalOperation: true,
         ),
         new CQRSCreate(
-            uriTemplate: '/search-aliases',
+            uriTemplate: '/search-aliases/{searchTerm}',
```
jolelievre: I'm thinking, with the normalizer now available, do we still need to keep this in the URI?

It was initially a little hack to get the `searchTerm` but we could use the normalizer to get this one from the JSON body as well no? This would allow keeping the same convention that POST request don't include the identifier in their URI

tests/Integration/ApiPlatform/SearchAliasEndpointTest.php
```diff
@@ -103,12 +103,36 @@ public function testCreateSearchAlias(): void
...
+                    'alias' => 'summer-clothing',
+                    'enabled' => true,
+                ],
+                [
+                    'alias' => 'summer-garment',
+                    'enabled' => false,
+                ],
+            ],
+        ];
+        $this->createItem('/search-aliases/summer%20dress', $postData, ['search_alias_write'], Response::HTTP_CREATED);
```
jolelievre: The JSON return of the method should also be tested, in the first call it should have limited aliases, and when you call it again we should see that they have been merged with the previous values Thus we can test and validate the cumulative effect of this particular endpoint

tests/Integration/ApiPlatform/SearchAliasEndpointTest.php
```diff
@@ -161,7 +185,7 @@ public function testUpdateSearchOnlyAliases(): void
                 ],
             ],
         ];
-        $this->updateItem('/search-aliases/' . self::SEARCH_TERM, $updateData, ['search_alias_write'], Response::HTTP_NO_CONTENT);
+        $this->updateItem('/search-aliases/' . self::SEARCH_TERM, $updateData, ['search_alias_write'], Response::HTTP_OK);
```
jolelievre: Similarly here, the return should also be tested But actually this modification should be done in all the tests of this class It wasn't possible until then because we couldn't properly return the whole entity

Now that it has been fixed the tests can be done correctly according to the rules we want to enforce on other endpoints:
- test the return
- test it as a whole array not partially

tests/Integration/ApiPlatform/SearchAliasEndpointTest.php
```diff
@@ -394,37 +437,12 @@ public function testGetNonExistentSearchAlias(): void
 
     public function testCreateSearchAliasWithInvalidData(): void
     {
-        // Test with empty search term - should return 422 with validation errors
-        $postData = [
```
jolelievre, line 398: Considering the comment I made above (I think we can remove the searchTerm from the POST uri), this should be testable again

jolelievre (changes requested):
Nice, but I think thanks to the normalizer a few things can be improved (some compromise we made that are not needed anymore)

jolelievre (approved):

ps-jarvis:
PR merged, well done!

PR #161, opened 2026-03-18
Fix bulk error status code

### LINKED ISSUE #40527: Admin API: handle bulk exception error messages better
When we have an error during bulk action (ex: I try to enabled via bulk actions but some of the IDs are non existent)
The current JSON returned is confusing and doesn't give the details of the faulty error

We need an error handler (in the core) that is able to correctly interpret the bulk exception, can loop through all the exceptions it contains and add an array of error messages on the JSON response (to be defined)

boherm:
| Questions         | Answers
| Description?      | Related to https://github.com/PrestaShop/PrestaShop/pull/41009.<br>Fix the bulk status code.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/PrestaShop/issues/40527
| How to test?      | CI ✅ + API Tests ✅

### DIFF
--- a/tests/Integration/ApiPlatform/ApiTestCase.php
+++ b/tests/Integration/ApiPlatform/ApiTestCase.php
@@ -285,6 +285,25 @@ abstract class ApiTestCase extends SymfonyApiTestCase
         return $this->requestApi(Request::METHOD_DELETE, $endPointUrl, $data, $scopes, $expectedHttpCode, $requestOptions);
     }
 
+    protected function bulkCommandItemsWithExpectedErrors(string $httpMethod, string $endPointUrl, ?array $data, ?array $expectedErrors, array $scopes = [], ?array $requestOptions = null): array|string|null
+    {
+        // In 9.1, we added a new normalizer to handle bulk command exceptions and return a 207 code with details about each error.
+        // If the normalizer is not present, it means we are on an older version and the API will return a 422 code with a generic error message,
+        // so we adapt our assertions based on the normalizer presence.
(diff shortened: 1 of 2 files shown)


tests/Integration/ApiPlatform/ApiTestCase.php
```diff
@@ -285,6 +285,22 @@ protected function bulkDeleteItems(string $endPointUrl, ?array $data, array $sco
         return $this->requestApi(Request::METHOD_DELETE, $endPointUrl, $data, $scopes, $expectedHttpCode, $requestOptions);
     }
 
+    protected function bulkCommandItemsWithExpectedErrors(string $httpMethod, string $endPointUrl, ?array $data, array $scopes = [], ?array $expectedErrors = null, ?array $requestOptions = null): array|string|null
```
jolelievre: Since the purpose of this method is to test only bulk endpoints when they fail, I don't think the `$expectedErrors` should be optional

tests/Integration/ApiPlatform/ApiTestCase.php
```diff
@@ -285,6 +285,22 @@ protected function bulkDeleteItems(string $endPointUrl, ?array $data, array $sco
...
+    {
+        $newNormalizer = class_exists("PrestaShopBundle\ApiPlatform\Normalizer\BulkCommandExceptionNormalizer");
+        $expectedHttpCode = $newNormalizer ? Response::HTTP_MULTI_STATUS : Response::HTTP_UNPROCESSABLE_ENTITY;
```
jolelievre, line 294: Maybe you can comment and explain this switch? Based on the difference between 9.0 and 9.1

jolelievre (approved):

PR #198, opened 2026-04-22
Fix execution file JSON parsing in validation step

jolelievre:
| Questions         | Answers
| Description?      | Fix the validation step that was failing because the execution output file is a JSON array, not a single object
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Related to PrestaShop/PrestaShop#41245
| How to test?      | 1. Merge this PR 2. Trigger the Claude pre-review on any PR 3. Verify the validation step completes and the execution report table is displayed in the step summary

## Details

The `claude-code-action` execution output file is a JSON array containing the full conversation log. The result object (with `is_error`, `num_turns`, `total_cost_usd`, etc.) is the **last element** of that array.

The previous code used `jq -r '.is_error'` directly on the file, which failed with: `jq: error: Cannot index array with string "is_error"` (exit code 5).

Fixed by extracting the last element first with `jq '.[-1]'` before reading individual fields.

### DIFF
--- a/.github/workflows/ai-prereview.yml
+++ b/.github/workflows/ai-prereview.yml
@@ -106,8 +106,14 @@ jobs:
             echo "::error::Claude execution output file not found."
             exit 1
           fi
-          # The execution file is a JSON array — the result object is the last element
-          RESULT=$(jq '.[-1]' "$EXECUTION_FILE")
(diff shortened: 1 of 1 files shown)


ps-jarvis:
PR merged, well done!

PR #201, opened 2026-04-30
Feat: add Cart Endpoint

axel-paillaud:
| Questions         | Answers
| Description?      | Add Cart Domain
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N A
| Related PR | https://github.com/PrestaShop/PrestaShop/pull/41364
| Sponsor company   | axel-paillaud
| How to test?      | Get a token, add cart_write and cart_read scope, test all endpoints with tools like cURL or Bruno

All endpoints require `Authorization: Bearer {token}` and `?shopId=1`.

| Method | Endpoint | Body |
|--------|----------|------|
| `POST` | `/carts` | `{"customerId": 1}` |
| `GET` | `/carts/{cartId}` | |
| `DELETE` | `/carts/{cartId}` | |
| `GET` | `/carts/{cartId}/view` | |
| `GET` | `/carts/last-empty/{customerId}` | |
| `POST` | `/carts/{cartId}/products` | `{"productId": 1, "quantity": 2, "combinationId": null, "customizationsByFieldIds": {}}` |
| `POST` | `/carts/{cartId}/products` (with customization) | `{"productId": 19, "quantity": 1, "combinationId": null, "customizationsByFieldIds": {"3": "My custom text"}}` |
| `DELETE` | `/carts/{cartId}/products/{productId}` | |
| `PATCH` | `/carts/{cartId}/products/quantity` | `{"productId": 1, "quantity": 3, "combinationId": null, "customizationId": null}` |
| `PATCH` | `/carts/{cartId}/products/price` | `{"productId": 1, "combinationId": 0, "price": 19.99}` |
| `POST` | `/carts/{cartId}/cart-rules` | `{"cartRuleId": 1}` |
| `DELETE` | `/carts/{cartId}/cart-rules` | `{"cartRuleId": 1}` |
| `POST` | `/carts/{cartId}/customizations` | `{"productId": 19, "customizationValuesByFieldIds": {"3": "My custom text"}}` |
| `PATCH` | `/carts/{cartId}/addresses` | `{"deliveryAddressId": 2, "invoiceAddressId": 2}` |
| `PATCH` | `/carts/{cartId}/carrier` | `{"carrierId": 2}` |
| `PATCH` | `/carts/{cartId}/currency` | `{"currencyId": 1}` |
| `PATCH` | `/carts/{cartId}/language` | `{"languageId": 1}` |
| `PATCH` | `/carts/{cartId}/delivery-settings` | `{"allowFreeShipping": false, "gift": false, "recycledPackaging": false, "giftMessage": null}` |
| `POST` | `/carts/{cartId}/send-email` _(deprecated)_ | |
| `DELETE` | `/carts/bulk-delete` | `{"cartIds": [2, 3]}` |

This PR need [#41364](https://github.com/PrestaShop/PrestaShop/pull/41364) to work properly, otherwise the customerId field is null.

Screenshot_2026-04-30_06-13-36

This doesn't prevent the endpoint from functioning correctly, but it shouldn't be null.


ps-jarvis:
Hello @axel-paillaud!

kpodemski:
Hello @axel-paillaud 

Thank you for contributing! Please find the AI review above, which should help you improve the pull request.

src/ApiPlatform/Resources/Cart/CartSendEmail.php
```diff
@@ -0,0 +1,52 @@
...
+
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ApiResource;
+use PrestaShop\PrestaShop\Core\Domain\Cart\Command\SendCartToCustomerCommand;
+use PrestaShop\PrestaShop\Core\Domain\Cart\Exception\CartNotFoundException;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSCreate;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * @deprecated Since PrestaShop 9.0. Will be removed in the next major version.
```
Hlavtox: Why are you implementing a deprecated endpoint? :D

Hlavtox:
@kpodemski Regarding the `SendCartToCustomerCommand`. It was replaced by `SendProcessOrderEmail`, but I think SendCartToCustomer is actually a better naming.

src/ApiPlatform/Resources/Cart/CartSendEmail.php
axel-paillaud: lol okay, removed

axel-paillaud:
> Hello @axel-paillaud
> 
> Thank you for contributing! Please find the AI review above, which should help you improve the pull request.

Hello @kpodemski,

The first point : `CartProductPrice.php: public float $price — CI-enforced hard blocker` is fixed : use DecimalPrice instead of float for price.

For the second point : `BulkDeleteCarts.php — $cartIds missing #[ApiProperty(identifier: true)]` : 

I suspect this is an error. Other bulk actions don't implement it (e.g., BulkDeleteZones), and `identifier: true` marks a unique instance of the resource (e.g., `$cartId` in `GET /carts/{cartId}`). For a bulk delete, there's no unique identifier; the URL is fixed (`carts/bulk-delete`). Do you confirm ?

axel-paillaud:
`#[LocalizedValue] / #[DefaultLanguage] — N/A, no localised fields` : 

Confirmed: no localised fields in the Cart resources. Product names and carrier names come from existing domain objects and are not stored directly in the Cart resource properties.

axel-paillaud:
**Cart.php — no CQRSQueryMapping for GetCartForOrderCreation; field alignment unverified**
Field alignment verified, all CartForOrderCreation getters map directly to Cart resource properties without renaming. No CQRSQueryMapping needed.

**CQRSQuery absent on CQRSCreate/CQRSPartialUpdate in most sub-resource classes**
Apparently, the other "action-oriented" endpoints (often patch or bulk) also don't use CQRSQuery. Do you want to use CQRSQuery for all Cart sub-resources, and therefore return the entire cart with each response, or do we leave it as is (returning only what was submitted)?

axel-paillaud:
**Missing validationContext on ALL 7 CQRSPartialUpdate operations**
**Missing validationContext on CQRSCreate in CartCartRules, CartCustomization, CartSendEmail**

Our Cart sub-resources have no constraints with `groups: ['Create']` or `groups: ['Update']`, so the absence of `validationContext` has no functional impact.

Other endpoints, such as `Profile`, `CustomerGroup`, and `Title`, also omit `ValidationContext` when validation groups are not differentiated.

axel-paillaud:
**CartProducts.php: uses 'Add' group instead of 'Create'**

fixed

axel-paillaud:
**Multi-shop**

**Shop context ([_context][shopId]/[_context][shopConstraint]) not forwarded to any CQRS command/query despite ?shopId=1 being required per PR description**

`?shopId=1` is required for API client authentication in multi-store mode, but the Cart domain has no ShopConstraint support, `CreateEmptyCustomerCartCommand` resolves the shop implicitly via `$customer->id_shop`. Forwarding the shop context to CQRS would require a Core domain change outside the scope of this PR.

axel-paillaud:
**Integration test**

 **testInvalidData only covers testCreateCartInvalidData; no validation tests for PATCH/sub-resource endpoints**
fixed

 **PATCH test methods do not assert response body/structure**
fixed

**testAddProductToCart asserts 'customization' => null in product structure, but Cart.php $products OpenAPI context defines no customization key — mismatch to clarify**
fixed

axel-paillaud:
@kpodemski I've just checked all the points raised by the AI, let me know if it's okay

tests/Integration/ApiPlatform/CartEndpointTest.php
```diff
@@ -0,0 +1,538 @@
...
+        parent::tearDownAfterClass();
+        self::resetTables();
+    }
+
+    protected static function resetTables(): void
+    {
+        DatabaseDump::restoreTables([
+            'cart',
+            'cart_product',
+            'cart_cart_rule',
```
mattgoud, line 59: the customization endpoints write to `customization` and `customized_data`, but they're not restored here — please add them so the cart tests don't leak state into other test classes:
```php
'cart',
'cart_product',
'cart_cart_rule',
'customization',
'customized_data',
```

tests/Integration/ApiPlatform/CartEndpointTest.php
```diff
@@ -0,0 +1,538 @@
...
+    /**
+     * @depends testCreateCart
+     */
+    public function testGetCart(int $cartId): int
+    {
+        $cart = $this->getItem('/carts/' . $cartId, ['cart_read']);
+
+        $this->assertEquals([
+            'cartId' => $cartId,
+            'customerId' => self::FIXTURE_CUSTOMER_ID,
```
mattgoud, line 191: this asserts `customerId` on the GET response, but `CartForOrderCreation` has no `customerId` getter so it's `null` until PrestaShop/PrestaShop#41364 is merged — that's the dependency noted in the PR description. this assertion (and the PR) is effectively blocked on that core PR landing first.

src/ApiPlatform/Resources/Cart/CartProductPrice.php
```diff
@@ -0,0 +1,68 @@
...
+    #[Assert\NotBlank]
+    #[Assert\Positive]
+    #[ApiProperty(openapiContext: ['type' => 'integer', 'example' => 1])]
+    public int $productId;
+
+    #[Assert\NotBlank]
+    #[ApiProperty(openapiContext: ['type' => 'integer', 'example' => 0])]
+    public int $combinationId;
+
+    #[Assert\NotBlank]
```
mattgoud: `#[Assert\NotBlank]` on a `DecimalNumber` is a no-op (an object is never "blank"). the other resources expose `DecimalNumber` with no validator — safe to drop this one. (non-blocking)

mattgoud (dismissed):
thanks for the PR @axel-paillaud 🙏 reviewing this as part of the sheriff rotation (AI review). this is a big, clean contribution — all 15 command mappings (param names, nullables, `COMMAND_MAPPING`) check out, the `CartView` query mapping is correct, the 18 endpoints are all covered by `getProtectedEndpoints`, and URIs/scopes/exceptions follow the conventions. a couple of blockers and some polish:

**1. blocked on core PR #41364 (not merged yet)**
the test asserts `customerId` in the cart responses (l.164 POST, **l.186 GET**, l.366), but `CartForOrderCreation` has no `customerId` getter, so it stays `null` on GET — exactly the dependency you documented. PrestaShop/PrestaShop#41364 is still open (and targets `9.1.x`), so the test will fail in CI against the current core. this PR can't go green until #41364 lands first. worth getting that one reviewed/merged as the prerequisite.

**2. `resetTables()` is missing the customization tables**
it only restores `cart`, `cart_product`, `cart_cart_rule`, but the add-customization tests (`/carts/{id}/customizations`) write to `customization` and `customized_data`, which never get restored → state leaks into other test classes. please add them.

**polish (non-blocking):**
- `CartProductPrice::$price` (`DecimalNumber`) has an `#[Assert\NotBlank]` that's a no-op on an object — the other resources expose `DecimalNumber` without a validator. safe to drop.
- small inconsistency: `CartCartRules` / `CartCustomization` POST omit `validationContext: ['groups' => ['Default', 'Create']]` and use ungrouped `NotBlank`, while `Cart` / `CartProducts` use the grouped pattern. both work, just worth standardizing.
- the description lists a deprecated `POST /carts/{cartId}/send-email` that doesn't exist in the diff (no resource, no test) — probably leftover, worth cleaning up.

nice work overall — main thing is the #41364 dependency. 👍

tests/Integration/ApiPlatform/CartEndpointTest.php
axel-paillaud: Done

src/ApiPlatform/Resources/Cart/CartProductPrice.php
axel-paillaud: Done

axel-paillaud:
Thank you for the detailed analysis. Here is a point-by-point response to each item raised.

---

## ✅ Fixed

### `langId` should be `languageId` (naming convention)
Renamed `$langId` to `$languageId` in `Cart.php` and updated the integration test accordingly.

### Missing `QUERY_MAPPING` on `Cart.php`
Added `QUERY_MAPPING = ['[langId]' => '[languageId]']` and wired it via `CQRSQueryMapping` on both `CQRSGet` and `CQRSCreate` operations that use `GetCartForOrderCreation`.

### Missing `validationContext` on `CartCartRules` and `CartCustomization`
Added `validationContext: ['groups' => ['Default', 'Create']]` to the `CQRSCreate` operation in both files, consistent with every other resource in the module.

### `#[Assert\NotBlank]` on `DecimalNumber` is a no-op
Removed the `#[Assert\NotBlank]` annotation from `CartProductPrice::$price`. The constraint had no effect on an object type and was misleading.

### `customization` and `customized_data` missing from `DatabaseDump::restoreTables`
Added both tables to `resetTables()` in `CartEndpointTest` to prevent state leaking into other test classes.

### Tautological assertions in `testCreateCart`, `testGetCart`, `testGetCartForViewing`
Replaced self-referencing values (`$cart['currencyId'] => $cart['currencyId']`) with concrete fixture constants. Added `FIXTURE_CURRENCY_ID = 1` and `FIXTURE_LANGUAGE_ID = 1` so the assertions now verify real values, while keeping the full `assertEquals` structure as required by the project conventions.

---

## ✅ Already correct - no change needed

### Identifier uses domain name + Id suffix (`cartId`)
All Cart resource classes already use `$cartId` (and `$customerId` for `LastEmptyCustomerCart`). This was correct from the initial commit.

### `cartsIds` → `cartIds` in `BulkDeleteCarts.php`
`BulkDeleteCarts::$cartIds` was already correctly named.

### `CQRSPartialUpdate` on `CartCartRules` DELETE
The DELETE operation already uses `CQRSDelete`. No change required.

### Scope format: `{entity_snake_case}_read` / `_write`
All 18 scope strings across the Cart resources already follow this convention (`cart_read`, `cart_write`).

---

## ✅ Confirmed N/A

### `shopIds` absent
Investigated the `Cart` ObjectModel: its `$definition['associations']` contains only `cart_rows` - there is no shop association. `CreateEmptyCustomerCartCommand` takes no `ShopConstraint`. Cart uses a single shop context (via `id_shop`), not a multi-shop association like `Contact` or `Supplier`. No `shopIds` field is needed.

---

## ⚠️ Not fixed - intentional or out of scope

### `CQRSQuery` missing on 5 PATCH and 2 POST sub-resource operations
These operations intentionally return no body: all integration tests verify the result via a subsequent `GET /carts/{cartId}` call, not the write response. This is consistent with `CartProducts` and `CartProductQuantity` which follow the exact same pattern. If a body is expected in the future, a `CQRSQuery` can be added at that point. A `204 No Content` status could be made explicit, but that is a separate discussion.

### Four sub-resource endpoints without a happy-path test
We investigated adding tests for `PATCH /products/price`, `PATCH /carrier`, `POST /cart-rules`, and `POST /customizations`.

- `PATCH /products/price`: The underlying `UpdateProductPriceInCartCommand` validates its price as a PHP `float`, but the resource exposes a `DecimalNumber`. There is a type mismatch that causes a `422` on any valid request. This is a pre-existing bug that deserves its own issue rather than a workaround in the test.
- `POST /customizations`: Product ID 1 in the test database has no customization fields (`customizable = 0`), so a meaningful test would require fixture setup that is out of scope for this PR.
- `PATCH /carrier` and `POST /cart-rules`: Technically feasible, but @mattgoud did not flag these as blocking, and adding them would require non-trivial fixture setup (creating a `CartRule` via ObjectModel). This is better addressed in a dedicated test-coverage PR.

tleon (dismissed):

nicosomb (dismissed):

src/ApiPlatform/Resources/Cart/CartAddresses.php
```diff
@@ -0,0 +1,68 @@
...
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Validator\Constraints as Assert;
+
+#[ApiResource(
+    operations: [
+        new CQRSPartialUpdate(
+            uriTemplate: '/carts/{cartId}/addresses',
+            requirements: ['cartId' => '\d+'],
+            CQRSCommand: UpdateCartAddressesCommand::class,
+            scopes: ['cart_write'],
```
jolelievre: We should use CQRSQuery here so that the endpoint returns the modified state of the cart, to avoid duplicating all the fields this endpoint should probably be moved into the `Cart` APi resource so that all the fields are present to display the returned state

src/ApiPlatform/Resources/Cart/CartCarrier.php
```diff
@@ -0,0 +1,62 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Cart\Exception\CartConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Cart\Exception\CartNotFoundException;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSPartialUpdate;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Validator\Constraints as Assert;
+
+#[ApiResource(
+    operations: [
+        new CQRSPartialUpdate(
+            uriTemplate: '/carts/{cartId}/carrier',
```
jolelievre: Similar feedback as for the adresses, the full cart should be returned after modification so it should be mutualized in a single API resource class It would be nice to add a comment on the fields that are meaningful only for some write endpoints, and an even cleaner solution would be to use serialization groups so that the open API only lists the relevant fields for the input BODY structure

src/ApiPlatform/Resources/Cart/CartCartRules.php
```diff
@@ -0,0 +1,66 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Cart\Exception\CartNotFoundException;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSCreate;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSDelete;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Validator\Constraints as Assert;
+
+#[ApiResource(
+    operations: [
+        new CQRSCreate(
+            uriTemplate: '/carts/{cartId}/cart-rules',
```
jolelievre: Same feedback as above

src/ApiPlatform/Resources/Cart/CartCurrency.php
```diff
@@ -0,0 +1,62 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Cart\Exception\CartConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Cart\Exception\CartNotFoundException;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSPartialUpdate;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Validator\Constraints as Assert;
+
+#[ApiResource(
+    operations: [
+        new CQRSPartialUpdate(
+            uriTemplate: '/carts/{cartId}/currency',
```
jolelievre: Again, the same feedback as above

src/ApiPlatform/Resources/Cart/CartCustomization.php
```diff
@@ -0,0 +1,68 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Cart\Exception\CartNotFoundException;
+use PrestaShop\PrestaShop\Core\Domain\Product\Customization\Exception\CustomizationConstraintException;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSCreate;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Validator\Constraints as Assert;
+
+#[ApiResource(
+    operations: [
+        new CQRSCreate(
+            uriTemplate: '/carts/{cartId}/customizations',
```
jolelievre: Again same feedback as above

src/ApiPlatform/Resources/Cart/CartDeliverySettings.php
```diff
@@ -0,0 +1,73 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Cart\Exception\CartNotFoundException;
+use PrestaShop\PrestaShop\Core\Domain\Cart\Exception\InvalidGiftMessageException;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSPartialUpdate;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Validator\Constraints as Assert;
+
+#[ApiResource(
+    operations: [
+        new CQRSPartialUpdate(
+            uriTemplate: '/carts/{cartId}/delivery-settings',
```
jolelievre: Same feedback as above as well, besides the delivery settings are already available in the Cart API resource, except they are in a sub array so we should be able to use a write structure that is the same as the read one with proper mapping So the JSON input would look like:
```
{
   "deliverySettings": {
      "allowFreeShipping": ...,
      "gift": ...
   }
}
```
And since it's a partial update the `CQRSPartialUpdate` and PATCH method is a perfect match for this use case The other update endpoints are all updating only partial parts of the Cart so they also should rely on `CQRSPartialUpdate`

src/ApiPlatform/Resources/Cart/CartProductPrice.php
```diff
@@ -0,0 +1,67 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Cart\Exception\CartNotFoundException;
+use PrestaShop\Decimal\DecimalNumber;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSPartialUpdate;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Validator\Constraints as Assert;
+
+#[ApiResource(
+    operations: [
+        new CQRSPartialUpdate(
+            uriTemplate: '/carts/{cartId}/products/price',
```
jolelievre:
```suggestion
            uriTemplate: '/carts/{cartId}/products/{productId}/price',
```
I think this type of url would reflect more the fact we are targeting one sub part product of the cart (and always the same idea that it should be mutualized in a single API resource so we can return the whole cart after modification)

src/ApiPlatform/Resources/Cart/CartLanguage.php
```diff
@@ -0,0 +1,62 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Cart\Command\UpdateCartLanguageCommand;
+use PrestaShop\PrestaShop\Core\Domain\Cart\Exception\CartConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Cart\Exception\CartNotFoundException;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSPartialUpdate;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Validator\Constraints as Assert;
+
+#[ApiResource(
+    operations: [
+        new CQRSPartialUpdate(
```
jolelievre: To be moved in single API resource and have the endpoint return the whole cart

src/ApiPlatform/Resources/Cart/CartProductQuantity.php
```diff
@@ -0,0 +1,68 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Cart\Exception\CartConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Cart\Exception\CartNotFoundException;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSPartialUpdate;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Validator\Constraints as Assert;
+
+#[ApiResource(
+    operations: [
+        new CQRSPartialUpdate(
+            uriTemplate: '/carts/{cartId}/products/quantity',
```
jolelievre:
```suggestion
            uriTemplate: '/carts/{cartId}/products/{productId}/quantity',
```
The require productId can be in the URL as well like for price, since combinationId and customizationId are optional I think it makes sense to keep them in the JSON body or the url will get too complex

src/ApiPlatform/Resources/Cart/CartProductPrice.php
jolelievre: The combinationId should be optional (for products without combinations)

jolelievre: On second thought, having everything in the unique Cart api resource will probably make it messy However, we can mutualize all the endpoints related to product updates in a single CartProduct api resource
Too bad we don't have a dedicated query that returns only the Cart products, that said we could cheat a little and use `GetCartForViewing` or `GetCartForOrderCreation` as a CQRSQuery for all the product endpoints but we can only map the returned products
So when you modify/add/delete a product in a cart you get the updated list of products, does it make sense for you?

The endpoints that change information at the cart level can be moved into Cart; however

src/ApiPlatform/Resources/Cart/CartView.php
```diff
@@ -0,0 +1,67 @@
...
+class CartView
+{
+    #[ApiProperty(identifier: true)]
+    public int $cartId;
+
+    #[ApiProperty(openapiContext: ['type' => 'integer', 'example' => 1])]
+    public int $currencyId;
+
+    #[ApiProperty(openapiContext: ['type' => 'object'])]
+    public array $customerInformation;
```
jolelievre: We should detail the content of those flattened arrays more for the OpenAPI documentation

jolelievre (changes requested):
Hi @axel-paillaud 

Thanks for this big contribution, and sorry for the late review with request changes after the PR already had two approvals.

But I noticed a few things that should be improved, the endpoints we build for the admin API try to follow a REST convention and they usually return the updated data, most of the time it's easy for small entities because the input and output match and we can simply use the same query and mapping as for the GET endpoint

Of course the cart domain is much more complex and has many fields so it's harder, but I still think we can reach a solution that would be more in tune with this spirit: write endpoints display full JSON of the modified entity to show the udpated state

I made a few comments in that spirit in the whole PR, mutualizing everything in only one Api resource may not be the right approach, but we could probably limit to two resources for most of them:
- one dedicated for the cart configuration that returns everything
- one dedicated for the cart's products that would return only the list of updated products

What do you think of this suggestion? Do you see how this could be done/implemented? If you disagree with the approach or if you have any question please don't hesitate to comment

jolelievre:
Ah also one important detail: I noticed this PR depends on https://github.com/PrestaShop/PrestaShop/pull/41364
Does it mean the endpoints cannot work AT ALL on previous versions? Or does it "simply" mean the customerId will only work with 9.2+?

If it's acceptable that the endpoint can be used with this missing field, we can integrate the endpoints for older versions, if you think it's mandatory to have it then we can add the new extraProperty `minVersion` to clearly indicate the compatibility of the related endpoints (they will be filtered in previous PrestaShop versions). This is a recent new feature we added not long ago

axel-paillaud:
Hello @jolelievre, thanks a lot for the feedback. I’ll take the time to analyze it thoroughly.

> Ah also one important detail: I noticed this PR depends on [PrestaShop/PrestaShop#41364](https://github.com/PrestaShop/PrestaShop/pull/41364) Does it mean the endpoints cannot work AT ALL on previous versions? Or does it "simply" mean the customerId will only work with 9.2+?
> 

I can answer that right away: the endpoint works without this PR, it’s just that the `customerId` field returns 'null' instead of the correct `customerId`.

You can see the before-and-after screenshots in the PR.

jolelievre:
@axel-paillaud on then if we consider it can work without your fix no need to limit the endpoints to a specific min version But it means in the integration tests, you need to check the customerId value differently depending on the core version (we have new helpers, isVersionAtLeast to help implement this)

axel-paillaud:
Hello @jolelievre 
Thanks again for the review. I’ve just taken another look at the Cart endpoints.

I think you're right; there are too many API Resources, especially compared to the other endpoints already in place.

I propose the following breakdown (5 files instead of 15) :

## `Cart.php`

| Method | URI | Command / Query | Moved from |
|---|---|---|---|
| GET | `/carts/{cartId}` | `GetCartForOrderCreation` | `Cart.php` |
| POST | `/carts` | `CreateEmptyCustomerCartCommand` | `Cart.php` |
| DELETE | `/carts/{cartId}` | `DeleteCartCommand` | `Cart.php` |
| PATCH | `/carts/{cartId}/addresses` | `UpdateCartAddressesCommand` | `CartAddresses.php` |
| PATCH | `/carts/{cartId}/carrier` | `UpdateCartCarrierCommand` | `CartCarrier.php` |
| PATCH | `/carts/{cartId}/currency` | `UpdateCartCurrencyCommand` | `CartCurrency.php` |
| PATCH | `/carts/{cartId}/language` | `UpdateCartLanguageCommand` | `CartLanguage.php` |
| PATCH | `/carts/{cartId}/delivery-settings` | `UpdateCartDeliverySettingsCommand` | `CartDeliverySettings.php` |
| POST | `/carts/{cartId}/cart-rules` | `AddCartRuleToCartCommand` | `CartCartRules.php` |
| DELETE | `/carts/{cartId}/cart-rules` | `RemoveCartRuleFromCartCommand` | `CartCartRules.php` |
| PUT | `/carts/{cartId}/emails` | `SendCartToCustomerCommand` | `CartEmail.php` |

## `CartProduct.php`

| Method | URI | Command | Moved from |
|---|---|---|---|
| POST | `/carts/{cartId}/products` | `AddProductToCartCommand` | `CartProducts.php` |
| DELETE | `/carts/{cartId}/products/{productId}` | `RemoveProductFromCartCommand` | `CartProducts.php` |
| PATCH | `/carts/{cartId}/products/{productId}/quantity` | `UpdateProductQuantityInCartCommand` | `CartProductQuantity.php` |
| PATCH | `/carts/{cartId}/products/{productId}/price` | `UpdateProductPriceInCartCommand` | `CartProductPrice.php` |
| POST | `/carts/{cartId}/customizations` | `AddCustomizationCommand` | `CartCustomization.php` |

## Unchanged

| File | Reason |
|---|---|
| `CartView.php` | Read-only, uses a different query (`GetCartForViewing`). Same shape as `CustomerDetails.php` (`GET /customers/{customerId}/details`). |
| `LastEmptyCustomerCart.php` | Read-only, dedicated query. |
| `BulkDeleteCarts.php` | Consistent with `BulkDeleteCategories.php` / `BulkDeleteDiscounts.php`. |

One question here: should the customization endpoint follow the same pattern and become `/carts/{cartId}/products/{productId}/customizations`? It would be more consistent with the two endpoints above, since it is also scoped to a single product.

This follows the same pattern as other endpoints, like Category. What do you think?

src/ApiPlatform/Resources/Cart/CartCarrier.php
axel-paillaud: I see your point.

As far as I can tell, even if we group all the operations under `Cart` in our example, we still end up with separate schemas.

We can see this with the `Category` resource: it declares multiple operations in the same file but produces distinct schemas.

Example:

Exporting OpenAPI:
`php bin/console api:openapi:export > openapi.json`

Then, displaying the `Category` schemas (which are in the same file):
`jq -r '.components.schemas | keys[] | select(startswith("Category."))' openapi.json`

The output is:
Category.AddCategoryCommand
Category.EditCategoryCommand
Category.SetCategoryIsEnabledCommand

So, merging the cart endpoints into `Cart` and `CartProduct` should keep the OpenAPI input schemas scoped to each command, without needing `#[Groups]`.

I’d like to verify that this works for `Cart` as well. What do you think?

jolelievre: Regarding your comment about category, that's because the OpenAPI schemas are automatically generated for inputs by scanning the CQRS command (over which we have a few adapters to automatically adapt them to mapping and other stuff, but that's the general idea).

So yes, even with mutualized classes, the input/output schema adapts partially automatically, but in case the automatic build isn't enough the groups are sometimes useful as well. Especially for validation when we add constraints on the fields.

You can try indeed; I suggested two main classes, but maybe I missed some stuff and other classes may be relevant because there's another completely different format (like for lists, it makes sense to have a dedicated simple one).

jolelievre:
@axel-paillaud the split you described in your last comment makes sense to me, let's try it this way.

Regarding the unchanged, I also agree, except for `LastEmptyCustomerCart ` I don't really understand the interest of this query, in the core it's never used anywhere so it's probably a remanant of an old implementation that was abandonned but not removed. We should probably deprecate it in the core to remove it in next major.

In the Admin API I don't think it brings any value, so we should ignore this CQRS query and add it to the list of exceptions in `GenerateApiTrackingTableCommand::EXCLUDED_CQRS_CLASSES` so we remove it from our tracking as well.

Also regarding `/carts/{cartId}/products/{productId}/customizations` it makes perfect sense to include the productId in the url

BTW I just reviewed the depending core PR and relaunched some UI tests for final checking, if they're all green the PR should be merged soon, so it will unlock your tests for this PR

axel-paillaud:
Hello @jolelievre,

Here is my new proposal, with all the Cart resources in 4 files.

The generated openAPI documentation is like this:

```plaintext
Cart.AddCartRuleToCartCommand
Cart.CreateEmptyCustomerCartCommand
Cart.SendCartToCustomerCommand
Cart.UpdateCartAddressesCommand
Cart.UpdateCartCarrierCommand
Cart.UpdateCartCurrencyCommand
Cart.UpdateCartDeliverySettingsCommand
Cart.UpdateCartLanguageCommand
CartProduct.AddProductToCartCommand
CartProduct.UpdateProductPriceInCartCommand
CartProduct.UpdateProductQuantityInCartCommand
```

Each one only lists the fields of its own command, not the whole cart:

```json
"Cart.UpdateCartCarrierCommand": { 
"type": "object", 
"properties": { 
"cartId": { "$ref": "#/components/schemas/CartId" }, 
"carrierId": { "type": "integer" } 
}
}
```

A few fields are missing from the generated input schemas (`quantity`, the delivery settings fields). It's a pre-existing schema builder issue, not specific to Cart. I can open a separate issue.

What do you think?

(I will be on vacation tomorrow and back on Tuesday, September 1st, for any necessary corrections)

jolelievre:
FYI @axel-paillaud the core PR has just been merged https://github.com/PrestaShop/PrestaShop/pull/41364
So you should be able to test this extra field (but only after 9.2, the test should adapt dynamically depending on the core version)

jolelievre:
@axel-paillaud regarding your comment, indeed for each class input a schema is generated So if the whole domain is split into multiple commands the API endpoints should naturally follow the same logic Even if the endpoints are shared on the same API resource class their respective schemas are adapted based on the associated CQRS command, it also interprates the command mapping and existing fields in the API resource class so that the whole schema remains consistent

In theory, it's a good thing because it means its schema actually reflects the relevant fields that will be interpreted by each schema. These represent the input schema not the output one, so if they are not the same it's ok as long as it matches the expected endpoint contract.

Now maybe there are some bugs in the automatic build of these schemas; did you notice some actual inconsistencies that need to be repaired? (If it's a bug in the core, it won't block this PR) 

For the read endpoint, which also should have its own schema (the output one), you should find all the fields that are present in the query result object, and that also should reflect a consistent schema representing what is actually returned by the GET endpoint.

So if some specific fields are only meaningful for one update endpoint it makes sense if they are filtered out from other endpoints where they are neither displayed for read nor intepreted for write.

src/ApiPlatform/Resources/Cart/CartView.php
```diff
@@ -0,0 +1,148 @@
...
+            'products' => [
+                'type' => 'array',
+                'items' => [
+                    'type' => 'object',
+                    'properties' => [
+                        'id' => ['type' => 'integer'],
+                        'name' => ['type' => 'string'],
+                        'attributes' => ['type' => 'string'],
+                        'reference' => ['type' => 'string'],
+                        'supplier_reference' => ['type' => 'string'],
```
jolelievre: Ok, I know this is gonna be a bit of a pain to handle, but the whole JSON contract should be camel cased And ideally also match the same naming as the other GET endpoint (when possible with same values of course)

For the other GET endpoint having camel case was easy because the QueryResult was based on PHP objects that follow this convention. Here it's more complicated because the requsult is mostly compose of flatten arrays that all use snake case fields

But it should be doable using the `@index` feature we have for the mapping like `'[cartSummary][products][@index][supplier_reference]' => '[products][@index][supplierReference]',`

The trick here is that the mapping doesn't remove the original data, so if we used `'[cartSummary][products][@index][supplier_reference]' => '[cartSummary][products][@index][supplierReference]',` we would end up with the same cartSummary that contains both snake case and camel case. But if we add another `public array $products;` field (that doesn't exist in the original QueryResult and normalized data), I believe it should work because we build the $products index by index, and then we don't include the $cartSummary so the original snake case format is filtered out from the final ApiResource.

I'm not 100% sure about this approach, and clearly it's a hack 😅 But it's worth trying so we can respect our global convention to use camel case only to maintain a homogenous contract.

Ideally you should inspire from the structure in the `Cart` ApiResource so we try and maintain a similar contract/format

jolelievre: And by extension this principle should be applied for all the other information in this `cartSummary` (cart rule + cart fields) so everything is nicely displayed in camel case

src/ApiPlatform/Resources/Cart/Cart.php
```diff
@@ -0,0 +1,348 @@
...
+            CQRSCommand: RemoveCartRuleFromCartCommand::class,
+            scopes: ['cart_write'],
+            output: Cart::class,
+            extraProperties: [
+                'CQRSQuery' => GetCartForOrderCreation::class,
+                'CQRSQueryMapping' => self::QUERY_MAPPING,
+            ],
+        ),
+        new CQRSUpdate(
+            uriTemplate: '/carts/{cartId}/emails',
```
jolelievre, line 145:
```suggestion
            uriTemplate: '/carts/{cartId}/send-cart-by-email',
```
This is a particular endpoint that does a very specific action So the endpoint should reflect this action so we understand it. Of course the rules don't allow using singular forms, that's why in the `ApiResourceUriTemplateRector` we added a `SKIPPED_KEYWORDS` array of accepted exception We use it to add the actions we need in singular, most of them are generic (toggle, uppload, reset, ...) But we can also add specific ones there (like send-password-reset-email)

src/ApiPlatform/Resources/Cart/Cart.php
```diff
@@ -0,0 +1,348 @@
...
+            CQRSQueryMapping: self::QUERY_MAPPING,
+            scopes: ['cart_write'],
+        ),
+        new CQRSDelete(
+            uriTemplate: '/carts/{cartId}',
+            requirements: ['cartId' => '\d+'],
+            CQRSCommand: DeleteCartCommand::class,
+            scopes: ['cart_write'],
+        ),
+        new CQRSPartialUpdate(
```
jolelievre, line 72: Hmm, it's a tricky opinion but I think we should use a PUT endpoint here:

```suggestion
        new CQRSUpdate(
```

I'll explain my opinion but it's debatable. In theory we are only sending partial data if we think in a cart perspective (in this case only addresses), then we return the whole cart. But we are not using an endpoint for a singular cart which would be `/carts/{cartId}`, we are using a partial endpoint mapping a bigger entity. Furthermore the command used beind doesn't accept partial data we MUST provide both delivery AND invoice address in each call of this endpoint, so the whole data handled by the endpoint must be provided in each call, and this type of enpoint matches more the PUT convention for rest (full data).

That's why for this endpoint (and the other "partial endpoints") I'd recommend using PUT on all of them when all the parameters must be provided, this makes the contract clearer and indicate you can't PUT only one of the fields, they are all required.

Of course it's far from ideal from a user/developer perspective to have all those endpoints split but we are restricted by the way the CQRS commands were created a long time ago, hence the multiple endpoints for each individual partial part.

I think in the future it would be interesting to create a new CQRS command that allows updating the whole cart with partial setters, that will allow us to have a much simpler PATCH endpoint for the whole entity. But it can only be added in a future version, so in the meantime we have to try the best with what we have at our disposal.

jolelievre: Of course, if you disagree with this approach, I'm open to discussion, and I'd be happy to hear about your opinion and arguments.

src/ApiPlatform/Resources/Cart/CartProduct.php
```diff
@@ -0,0 +1,164 @@
...
+            uriTemplate: '/carts/{cartId}/products/{productId}',
+            requirements: ['cartId' => '\d+', 'productId' => '\d+'],
+            // CQRSDelete answers 204 and exposes no CQRSQuery parameter, hence the overrides.
+            status: Response::HTTP_OK,
+            CQRSCommand: RemoveProductFromCartCommand::class,
+            scopes: ['cart_write'],
+            output: CartProduct::class,
+            extraProperties: ['CQRSQuery' => GetCartForOrderCreation::class],
+        ),
+        new CQRSPartialUpdate(
```
jolelievre, line 64: Similar to my comment above about PUT method I think it could be applied here as well. So yes some fields are not required (combinationId and customizationId) but it's not because we can send partial data it's because they are not part of the identification of the product/combination/customization we want to update.

However the data to send the price is always required (and it's the most relevant data anyway since it's the only one that updates a value).

Although on this one I'm less sure of this approach 😅 But if we go in that direction for the endpoints detailed above it would at least maintain some consistency on the whole cart domain endpoints.

src/ApiPlatform/Resources/Cart/CartProduct.php
```diff
@@ -0,0 +1,164 @@
...
+    #[ApiProperty(openapiContext: ['type' => 'integer', 'example' => null, 'nullable' => true])]
+    public ?int $combinationId;
+
+    // Optional, to target a single customized line of the product rather than all of them
+    #[Assert\Positive(groups: ['UpdateQuantity'])]
+    #[ApiProperty(openapiContext: ['type' => 'integer', 'example' => null, 'nullable' => true])]
+    public ?int $customizationId;
+
+    // Optional, text customizations to attach to the product being added. Only text fields are supported: file
+    // fields expect an UploadedFile, which a JSON body cannot carry.
```
jolelievre, line 156: Hmm, in theory we should be able to handle the file upload since the command handles it But let's postpone it for now because the PR is already complex enough
It can be improved later

src/ApiPlatform/Resources/Cart/CartView.php
jolelievre: Similarly if we want to reformat `customerInformation` into camel case we can create a new `public array customer;` in which we can build the object with camel case fields

tests/Integration/ApiPlatform/CartEmailEndpointTest.php
```diff
@@ -59,13 +59,17 @@ public function testSendCartToCustomer(): void
...
             'PUT',
             '/carts/' . $cartId . '/emails',
             null,
             ['cart_write'],
-            Response::HTTP_NO_CONTENT
+            Response::HTTP_OK
         );
+
+        // Like every other Cart write, the endpoint answers with the cart itself
+        $this->assertEquals($cartId, $cart['cartId']);
```
jolelievre, line 71: I didn't notice the specificity of this one, it's a particular action to send an email and it doesn't update anything in the cart content when we perform the action So this one could return an empty response no content

But since you implemented the body response I won't block in either case anyway, I leave the decision to you regarding the most relevant approach for this case

tests/Integration/ApiPlatform/CartEndpointTest.php
```diff
@@ -0,0 +1,561 @@
...
+        ], $cart);
+
+        return $cartId;
+    }
+
+    /**
+     * Standalone on purpose: CartForOrderCreation only exposes the customer from 9.2 on
+     * (PrestaShop/PrestaShop#41364), and a failure here must not skip the whole cart flow below.
+     */
+    public function testCartExposesCustomerId(): void
```
jolelievre: I'm not sure you need a dedicated test for this, instead you can include it in your expected data in `testGetCart` or in any other test functions. It's just that you'll have to dynamise the expected data because it's only available starting from 9.2:

```php
$expectedData = [
    'cartId' => $cartId,
    'currencyId' => self::FIXTURE_CURRENCY_ID,
];
if (self::isVersionAtLeast('9.2.0')) {
    $expectedData['customerId'] = $cart['customerId'];
}
```

tests/Integration/ApiPlatform/CartEndpointTest.php
```diff
@@ -0,0 +1,561 @@
...
+        $reloadedCart = $this->getItem('/carts/' . $cart['cartId'], ['cart_read']);
+        $this->assertSame(self::expectedCustomerId(), $reloadedCart['customerId']);
+
+        $this->deleteItem('/carts/' . $cart['cartId'], ['cart_write']);
+    }
+
+    /**
+     * @depends testGetCart
+     */
+    public function testAddProductToCart(int $cartId): int
```
jolelievre, line 197: I'm not sure why this test is failing in the CI because it seems logic I didn't test locally yet though But maybe now that your PR has been merged in the core and if you handle the dynamic expected data it will unlock the problem

tests/Integration/ApiPlatform/CartEndpointTest.php
```diff
@@ -0,0 +1,561 @@
...
+
+        return $cartId;
+    }
+
+    /**
+     * @depends testUpdateProductPriceInCart
+     */
+    public function testRemoveProductFromCart(int $cartId): int
+    {
+        $response = $this->deleteItem(
```
jolelievre, line 275: To make sure the endpoint doesn't remove all the products maybe it would be interesting to have at least two products, remove one and check that the other is still present

tests/Integration/ApiPlatform/CartEndpointTest.php
```diff
@@ -0,0 +1,561 @@
...
+     */
+    public function testUpdateProductQuantityInCart(int $cartId): int
+    {
+        $response = $this->partialUpdateItem(
+            '/carts/' . $cartId . '/products/' . self::FIXTURE_PRODUCT_ID . '/quantity',
+            ['quantity' => 5],
+            ['cart_write']
+        );
+
+        $this->assertNotEmpty($response['products']);
```
jolelievre, line 245: I know it's a pain, sorry I only have troublesome requests 😅, but ideally we should check the whole response format for each endpoint. So the whole expected cart data should be check in each test functions. But to make it easier and not duplicate the whole data each time maybe you can return the evolving expectedData flatten array between the test methods (similar to what we do with the $cartId). This way you only need to update the sub part that was modified and then you can use it to make a full `assertEquals`

jolelievre (commented):
Thanks you for the hard work @axel-paillaud 

This is far from being a simple domain, mostly because the CQRS implementation is very fragmented. This page was migrated a long time ago when we started using CQRS, we probably wouldn't architecture the commands the same way today.

So sorry for the difficulty, but you're doing a great job on it. As usual don't hesitate to ask for any help if you need it.

axel-paillaud:
Hello @jolelievre,

No worries about the corrections; I understand that this Cart resource is quite complex, and I appreciate the time you are taking to put together a proper pull request. It helps me improve as well.
The only issue is that this PR might take a little while, as my time each day is limited (like everyone else's).

## Regarding the apidocs documentation:

### Core-related bug:

#### PATCH /carts/{cartId}/products/{productId}/quantity: `quantity` is missing from the documented request body.

Generated schema: `cartId, combinationId, customizationId, productId`, `required: []`.
The main field is missing.

Cause: `UpdateProductQuantityInCartCommand` has a constructor parameter named `$quantity` but stores the value in the `$newQuantity` property (line 35).

swagger schema screenshot, without the main field, quantity:

Screenshot_2026-09-04_09-38-39

#### PATCH /carts/{cartId}/delivery-settings: documented body is empty.

Generated schema: `cartId` only. The `shipping` field is lost.

Cause: `SchemaSynchronizer.php:99` performs `trim($apiPath, '[]')`. On a nested path like `[shipping][freeShipping]`, this results in `shipping][freeShipping` instead of `shipping`. **Consequently, no nested `CQRSCommandMapping` entries can be recognized.** This is a generic core bug, not specific to the Cart resource.

### Corrections made in the Admin API (fixable here, not core-related):

(commit 7b07cae)

The read schema listed fields the endpoints never return: `carrierId`, `cartRuleId`, `deliveryAddressId` and `invoiceAddressId` on `Cart`, everything but `cartId` and `products` on `CartProduct`. They only exist as write inputs. I marked them `#[ApiProperty(readable: false)]`, which flags them `writeOnly` in the OpenAPI document so Swagger UI drops them from the response schema and example; the input schemas are unchanged. Serialization groups would work too (I think), but this was enough without introducing a convention unused elsewhere in the module. Happy to switch if you prefer.

src/ApiPlatform/Resources/Cart/CartView.php
axel-paillaud: I see your point. Here is my proposal, commit 9cd6d2a, which uses @index and rebuilds all fields via QUERY MAPPING to obtain camelCase.

Also, after testing: **nested `@index` placeholders don't work**

Everything is camelCase now, except one field. The `NormalizationMapper` cannot expand a second `@index` nested inside a first one, so `customization.fields` can't be mapped element by element:

```php
// silently ignored, no error, the field just disappears
'[cartSummary][products][@index][customization][fields][@index2][allow_html]'
    => '[products][@index][customization][fields][@index2][allowHtml]',
```

`updateMappingIndexes()` iterates over a copy of the mapping, so the entries created when expanding the first placeholder are never reprocessed for the second one. I worked around it by copying `customization` as a whole, which means `allow_html` stays snake_case, the only remaining exception in the contract.

Also worth knowing: when the source is an empty list, the target path is not created at all (no key in the response, not even `[]`). I gave the rebuilt properties default values to keep the contract stable. A fix looks small: loop `updateMappingIndexes()`, in `src/PrestaShopBundle/ApiPlatform/NormalizationMapper.php` until no placeholder is left.

Quetzacoalt91:
Hi @axel-paillaud, looks like the added code fails during the checks, can you have a look at it? Thanks

src/ApiPlatform/Resources/Cart/Cart.php
axel-paillaud: I agree with you. Since we send all the fields as input, even when changing just one, we are adhering to the PUT convention rather than PATCH.

I suggest applying this to `/delivery-settings` as well, to maintain consistency, even though it has two nullable fields and a PATCH approach could be justified.

src/ApiPlatform/Resources/Cart/Cart.php
axel-paillaud: Done. Note that PUT /carts/{cartId}/emails was shipped in v0.9.0 (from #314), so this rename is a BC break on a released endpoint. Fine by me on a 0.x, but flagging it in case you want it in the release notes.

tests/Integration/ApiPlatform/CartEmailEndpointTest.php
axel-paillaud: Agreed, it doesn’t change the cart so it now answers 204 with no body

tests/Integration/ApiPlatform/CartEndpointTest.php
axel-paillaud: Done: the test now adds a second product, removes the first one and asserts the second is still in the cart, then removes it too.

tests/Integration/ApiPlatform/CartEndpointTest.php
axel-paillaud: Done: the chained tests now pass an evolving $expectedCart instead of the cart ID, each one updates the part its endpoint modifies and asserts the whole response.

axel-paillaud:
Hello @Quetzacoalt91, sorry for the delay, and thanks for fixing the checks and iterating on the endpoints.

@jolelievre I’ve just fixed the last few outstanding points. Don't hesitate to reach out if you have any other suggestions!

PR #214, opened 2026-06-10
Add Category get-status endpoint

PrestaEdit:
| Questions         | Answers
| Description?      | Add a Category get-status endpoint to the Admin API
| BC breaks?        | no
| Deprecations?     | no
| Sponsor company   |

Exposes `GetCategoryIsEnabled` through **`GET /categories/{categoryId}/status`**, the read
counterpart of the existing `PATCH /categories/{categoryId}/status`. The query returns a
scalar boolean, mapped to the resource via `[_queryResult] => [enabled]` (same pattern as
the ShowcaseCard endpoint); the GET and PATCH share a path within the same resource class,
like the Hook `/status` operations.

Adds an integration test and a scopes (authorization) entry.


PrestaEdit:

PR #215, opened 2026-06-10
Add Tag resource

PrestaEdit:
| Questions         | Answers
| Description?      | Add the Tag resource (CRUD) to the Admin API
| BC breaks?        | no
| Deprecations?     | no
| Sponsor company   |

Adds the Admin API endpoints for the **Tag** domain, mirroring existing simple CRUD
resources (Title, SearchEngine):

- `POST /tags` — `AddTagCommand`
- `GET /tags/{tagId}` — `GetTagForEditing`
- `PATCH /tags/{tagId}` — `EditTagCommand`
- `DELETE /tags/{tagId}` — `DeleteTagCommand`
- `DELETE /tags/bulk-delete` — `BulkDeleteTagCommand`

A tag belongs to a single language, so the resource exposes `name` + `languageId`
(plus optional `productIds` on write, `products` on read). Resource property names match
the CQRS command/query fields, so no explicit mapping is needed (same as SearchEngine).

Adds an integration test covering the full CRUD + bulk delete, and the `tag_read` /
`tag_write` scopes.

### DIFF
--- a/src/ApiPlatform/Resources/Tag/BulkDeleteTags.php
+++ b/src/ApiPlatform/Resources/Tag/BulkDeleteTags.php
@@ -1,0 +1,56 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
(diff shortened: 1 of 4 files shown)


PrestaEdit:
Converting to draft: the Tag CQRS layer (`AddTagCommand`, `GetTagForEditing`, etc.) does not exist in PrestaShop **9.0.3**, which this module still supports (PHPStan 9.0.3 reports the classes as not found, and the 9.0.3 integration jobs fail accordingly). The 9.1.x jobs pass. This endpoint can only be added once 9.0.3 support is dropped, or behind a version guard. Holding until then.

kpodemski:
@PrestaEdit 

we can drop 9.0.3 support, this module hasn't reached the v1 just yet, so feel free to drop 9.0.x tests

unless we have a better way, @jolelievre ?

PrestaEdit:
@kpodemski I went ahead and opened #220 to drop 9.0.x from the CI matrices (PHPStan + integration), removing the dedicated `phpstan-9.0.3.neon` and the 9.0.3-only PHP 8.4/8.5 rules. I kept it CI-only for now — `ps_versions_compliancy['min']` still allows 9.0.x, so declared support is unchanged, only the test coverage is dropped.

Once #220 is merged this PR's 9.1-only Tag CQRS classes are fine, so I'll take it out of draft then. Happy to adjust if @jolelievre prefers a different approach.

kpodemski:
thanks @PrestaEdit

PR #216, opened 2026-06-10
Add OrderMessage resource

PrestaEdit:
| Questions         | Answers
| Description?      | Add the OrderMessage resource (CRUD) to the Admin API
| BC breaks?        | no
| Deprecations?     | no
| Sponsor company   |

Adds the Admin API endpoints for the **OrderMessage** domain, mirroring the existing
localized CRUD resource (Title):

- `POST /order-messages` — `AddOrderMessageCommand`
- `GET /order-messages/{orderMessageId}` — `GetOrderMessageForEditing`
- `PATCH /order-messages/{orderMessageId}` — `EditOrderMessageCommand`
- `DELETE /order-messages/{orderMessageId}` — `DeleteOrderMessageCommand`
- `DELETE /order-messages/bulk-delete` — `BulkDeleteOrderMessageCommand`

An order message has a localized `names` and `messages`, mapped to the command/query
`localizedName` / `localizedMessage` fields.

Adds an integration test covering the full CRUD + bulk delete, and the
`order_message_read` / `order_message_write` scopes.

### DIFF
--- a/src/ApiPlatform/Resources/OrderMessage/OrderMessage.php
+++ b/src/ApiPlatform/Resources/OrderMessage/OrderMessage.php
@@ -1,0 +1,103 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\OrderMessage;
+
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ApiResource;
+use PrestaShop\PrestaShop\Core\Domain\OrderMessage\Command\AddOrderMessageCommand;
+use PrestaShop\PrestaShop\Core\Domain\OrderMessage\Command\DeleteOrderMessageCommand;
+use PrestaShop\PrestaShop\Core\Domain\OrderMessage\Command\EditOrderMessageCommand;
+use PrestaShop\PrestaShop\Core\Domain\OrderMessage\Exception\OrderMessageConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\OrderMessage\Exception\OrderMessageNotFoundException;
+use PrestaShop\PrestaShop\Core\Domain\OrderMessage\Query\GetOrderMessageForEditing;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSCreate;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSDelete;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSGet;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSPartialUpdate;
+use PrestaShopBundle\ApiPlatform\Metadata\LocalizedValue;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Validator\Constraints as Assert;
+
+#[ApiResource(
+    operations: [
+        new CQRSCreate(
+            uriTemplate: '/order-messages',
+            validationContext: ['groups' => ['Default', 'Create']],
+            CQRSCommand: AddOrderMessageCommand::class,
+            CQRSCommandMapping: self::COMMAND_MAPPING,
+            scopes: ['order_message_write'],
+        ),
+        new CQRSDelete(
+            uriTemplate: '/order-messages/{orderMessageId}',
+            requirements: ['orderMessageId' => '\d+'],
+            output: false,
+            CQRSCommand: DeleteOrderMessageCommand::class,
+            scopes: ['order_message_write'],
+        ),
+        new CQRSGet(
+            uriTemplate: '/order-messages/{orderMessageId}',
+            requirements: ['orderMessageId' => '\d+'],
+            CQRSQuery: GetOrderMessageForEditing::class,
+            CQRSQueryMapping: self::QUERY_MAPPING,
+            scopes: ['order_message_read'],
+        ),
+        new CQRSPartialUpdate(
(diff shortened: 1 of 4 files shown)


tests/Integration/ApiPlatform/OrderMessageEndpointTest.php
```diff
@@ -0,0 +1,176 @@
...
+    {
+        parent::setUpBeforeClass();
+        DatabaseDump::restoreTables(['order_message', 'order_message_lang']);
+        self::createApiClient(['order_message_write', 'order_message_read']);
+    }
+
+    public static function tearDownAfterClass(): void
+    {
+        parent::tearDownAfterClass();
+        DatabaseDump::restoreTables(['order_message', 'order_message_lang']);
```
mattgoud, line 42: the entity is localized, so this should also reset languages: `LanguageResetter::resetLanguages();` (+ `use Tests\Resources\Resetter\LanguageResetter;`), like `TitleEndpointTest`. otherwise language state can leak into other test classes.

tests/Integration/ApiPlatform/OrderMessageEndpointTest.php
```diff
@@ -0,0 +1,176 @@
...
+
+        return $orderMessageId;
+    }
+
+    /**
+     * @depends testGetOrderMessage
+     */
+    public function testEditOrderMessage(int $orderMessageId): int
+    {
+        $updated = $this->partialUpdateItem('/order-messages/' . $orderMessageId, [
```
mattgoud, line 128: this only covers a `names`-only PATCH. could you add a `messages`-only case too? that path sends `localizedName = null` to the command and hits an unguarded `foreach(null)` in `EditOrderMessageHandler::assertNameIsNotAlreadyUsed()` (core) → warning/500. good to surface it here.

src/ApiPlatform/Resources/OrderMessage/OrderMessage.php
```diff
@@ -0,0 +1,103 @@
...
+            scopes: ['order_message_write'],
+        ),
+        new CQRSGet(
+            uriTemplate: '/order-messages/{orderMessageId}',
+            requirements: ['orderMessageId' => '\d+'],
+            CQRSQuery: GetOrderMessageForEditing::class,
+            CQRSQueryMapping: self::QUERY_MAPPING,
+            scopes: ['order_message_read'],
+        ),
+        new CQRSPartialUpdate(
```
mattgoud, line 64: since fields are optional on partial update, a PATCH omitting `names` reaches `EditOrderMessageHandler::assertNameIsNotAlreadyUsed()` which iterates `getLocalizedName()` (nullable) without a null guard. nothing to change in this file — flagging so it gets a companion core fix + test coverage.

mattgoud (changes requested):
thanks for the PR @PrestaEdit 🙏. The CQRS mappings, scopes, exception mapping and overall structure all check out and match the `Title` reference resource. a few things before merge:

**1. partial update with `messages` only isn't covered — and it reaches an unguarded path in core**
the `EditOrderMessageCommand` fields are nullable, so a `PATCH` sending only `messages` leaves `localizedName = null`. `handle()` guards that, but `EditOrderMessageHandler::assertNameIsNotAlreadyUsed()` does `foreach ($command->getLocalizedName() ...)` with no null check → `foreach(null)` (warning, turns into a 500 under the test/debug error handler). the test only covers a `names`-only PATCH so it stays hidden. could you add a "PATCH messages only" case? it's really a core bug (the guard belongs in the handler), but this endpoint is what makes it reachable, so worth a separate core fix too.

**2. missing `LanguageResetter` in teardown**
the entity has localized fields, so `tearDownAfterClass()` should call `LanguageResetter::resetLanguages()` like the other localized tests (`Title`, `Manufacturer`, `Discount`) — otherwise language state can leak across test classes.

**3. (non-blocking) no list endpoint**
the `Title` reference also ships a `TitleList` (paginated `GET /titles`). not required for this PR, but a `GET /order-messages` would round it out nicely if you're up for it.

nothing major — 1 & 2 and we're good 👍

PrestaEdit:
Thanks for the thorough review @mattgoud 🙏 — all three addressed:

**1. PATCH messages-only → unguarded core path.** You're right, it's a genuine core bug: `EditOrderMessageHandler::handle()` calls `assertNameIsNotAlreadyUsed()` unconditionally, and that does `foreach ($command->getLocalizedName() …)` → `foreach(null)` on a messages-only edit. I opened a core fix that guards the check (only run it when a name is actually provided, matching the existing null guards in `handle()`): **PrestaShop/PrestaShop#41790**.

I also added the `testEditOrderMessageMessagesOnly` case you asked for (PATCH only `messages`, assert names preserved). Heads-up: that test only goes green once #41790 is on the tested cores — until then it will be red on the released-core matrices, since the bug lives in core. Happy to sequence this however you prefer (hold the test until the core fix ships, or keep it here to track the dependency).

**2. `LanguageResetter` in teardown.** Added `LanguageResetter::resetLanguages()` to `tearDownAfterClass()`, like the `Title`/`Manufacturer`/`Discount` tests.

**3. List endpoint.** Added `GET /order-messages` (`OrderMessageList`, mirroring `TitleList` — `PaginatedList` + `QueryListProvider` + `OrderMessageFilters`) plus a `testListOrderMessages` case.

mattgoud (approved):
thanks for your contribution @PrestaEdit

SiraDIOP (commented):
Hello @PrestaEdit, 

Thanks for your PR 

GET /order-messages
The list endpoint returns 200 with the expected pagination fields (totalItems, orderBy, sortOrder, limit, filters, items). The existing "Delay" message is returned. Note that the list returns name and message (singular, non-localized) which seems intentional for the list view.

Image

GET /order-messages/1
Returns 200 with the full localized fields names and messages, both indexed by locale (en-US, fr-FR). Response structure is correct.

Image

POST /order-messages
Returns 201 with the newly created orderMessageId: 2. Both localized names and messages fields are accepted. Note: the response only returns the identifier, not the full resource — a second GET is needed to retrieve the created data. This is a known limitation flagged in the AI review (missing CQRSQuery on CQRSCreate).

Image

PATCH /order-messages/2 (names only)
Returns 200 with the updated names and the original messages preserved. Partial update works as expected.

Image

PATCH /order-messages/2 (messages only)
Returns 500. Sending only messages without names passes localizedName = null to EditOrderMessageCommand, which hits an unguarded foreach(null) in EditOrderMessageHandler::assertNameIsNotAlreadyUsed(). This is a known core bug tracked in PrestaShop/PrestaShop#41790. This test will pass once that fix is merged.

Image

DELETE /order-messages/2
Returns 204 No Content as expected.

Image

GET /order-messages/2 (after deletion)
Returns 404 with OrderMessageNotFoundException as expected. Minor inconsistency: the HTTP status code is correctly 404 but the body contains "status": 500 instead of 404. Worth fixing for consistency.

Image

DELETE /order-messages/bulk-delete
Returns 204 No Content as expected. Bulk delete works correctly.

Image

GET /order-messages/3 and /order-messages/4 (after bulk deletion)
Both return 404 as expected. Deleted resources are no longer accessible.

Image
Image

Thanks

SiraDIOP (approved):

ps-jarvis:
PR merged, well done!

Touxten:
Thank you @SiraDIOP && @PrestaEdit

PR #218, opened 2026-06-11
Add Customer private-note endpoint

PrestaEdit:
| Questions         | Answers
| Description?      | Add a Customer private-note endpoint to the Admin API
| BC breaks?        | no
| Deprecations?     | no
| Sponsor company   |

Exposes `SetPrivateNoteAboutCustomerCommand` through
**`PATCH /customers/{customerId}/private-notes`**, in a dedicated resource class so the
rich `Customer` resource is left untouched. The body carries the `privateNote` string
(mapped by matching field name). The URI uses the plural `private-notes` to satisfy the
module's `ApiResourceUriTemplateRector` convention.

Adds an integration test and a scopes (authorization) entry.

### DIFF
--- a/src/ApiPlatform/Resources/Customer/CustomerPrivateNote.php
+++ b/src/ApiPlatform/Resources/Customer/CustomerPrivateNote.php
@@ -1,0 +1,57 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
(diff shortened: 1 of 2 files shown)


PrestaEdit:
Superseded by #428, which merges this endpoint with the rest of the domain, following the mutualisation asked for by the core team (one PR per domain rather than one per endpoint).

PR #219, opened 2026-06-11
Add OrderReturnState resource

PrestaEdit:
| Questions         | Answers
| Description?      | Add the OrderReturnState resource (CRUD) to the Admin API
| BC breaks?        | no
| Deprecations?     | no
| Sponsor company   |

Adds the Admin API endpoints for the **OrderReturnState** domain, mirroring existing
localized CRUD resources:

- `POST /order-return-states` — `AddOrderReturnStateCommand`
- `GET /order-return-states/{orderReturnStateId}` — `GetOrderReturnStateForEditing`
- `PATCH /order-return-states/{orderReturnStateId}` — `EditOrderReturnStateCommand`
- `DELETE /order-return-states/{orderReturnStateId}` — `DeleteOrderReturnStateCommand`
- `DELETE /order-return-states/bulk-delete` — `BulkDeleteOrderReturnStateCommand`

An order return state has a localized `names` and a `color`. As with the Contact resource,
the create and update commands expect different field names (`localizedNames` vs `name`),
handled by two distinct command mappings.

Adds an integration test covering the full CRUD + bulk delete, and the
`order_return_state_read` / `order_return_state_write` scopes.

### DIFF
--- a/src/ApiPlatform/Resources/OrderReturnState/BulkDeleteOrderReturnStates.php
+++ b/src/ApiPlatform/Resources/OrderReturnState/BulkDeleteOrderReturnStates.php
@@ -1,0 +1,56 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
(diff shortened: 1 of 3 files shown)


boo-code:
Small consistency note on the localized field name: per [ADR 0023](https://github.com/PrestaShop/ADR/blob/master/0023-cqrs-api-guidelines.md) ("localized properties do not start by `localized`") and the established convention in this module — 26 of the 28 `#[LocalizedValue]` array properties across the resources use the unprefixed name, and `$names` specifically is already used by 4 other resources — `public array $localizedNames` here would be more consistent as `public array $names`.

Worth aligning before merge, since the property name is part of the API contract: renaming it after release is a BC break. (`CustomerGroup` already shipped with `$localizedNames`, so it's now BC-locked there — catching it here keeps the count from growing.) Minor, but essentially free to fix pre-merge.

PR #220, opened 2026-06-11
Drop 9.0.x from the CI test matrix

PrestaEdit:
| Questions         | Answers
| Description?      | Stop running the CI test suites against PrestaShop 9.0.x
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Follows up on #215 (comment)
| Sponsor company   |

As discussed on #215, the module has not reached v1 yet, so we can stop testing against PrestaShop 9.0.x.

This removes `9.0.3` from:
- the **PHPStan** matrix (`.github/workflows/php.yml`)
- the **integration** matrix (`.github/workflows/integration.yml`), along with the now-useless `9.0.3`-specific PHP `8.4` include and PHP `8.5` exclude rules
- the dedicated `tests/PHPStan/phpstan-9.0.3.neon` config (no longer referenced once `9.0.3` leaves the matrix)

Module installability is left unchanged on purpose: `ps_versions_compliancy['min']` still allows `9.0.x`. This PR only drops CI coverage, not declared support. Dropping the 9.0.x constraint also unblocks endpoints relying on CQRS classes introduced in 9.1 (e.g. the Tag domain in #215).


kpodemski (approved):

kpodemski:
cc @jolelievre for the second review

PR #231, opened 2026-06-15
Add Shop search endpoint

PrestaEdit:
| Questions         | Answers
| Description?      | Add a Shop search endpoint to the Admin API
| BC breaks?        | no
| Deprecations?     | no
| Sponsor company   |

Exposes `SearchShops` through **`GET /shops/search?searchTerm=...`**, returning the shop
groups and shops matching the term (the same data used by the back-office shop selector).
Implemented as a `CQRSGetCollection`, like the cart-rules search.

The query result mixes shop-group rows (`{id, color, name}`) and shop rows (which also carry
`groupId` / `groupName` / `groupColor`), represented by a single resource whose `group*`
fields are nullable.

Adds an integration test and the `shop_read` scope.

### DIFF
--- a/src/ApiPlatform/Resources/Shop/FoundShop.php
+++ b/src/ApiPlatform/Resources/Shop/FoundShop.php
@@ -1,0 +1,81 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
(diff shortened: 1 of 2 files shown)


PrestaEdit:
Superseded by #422, which merges this endpoint with the rest of the domain, following the mutualisation asked for by the core team (one PR per domain rather than one per endpoint).

PR #235, opened 2026-06-15
Add Currency resource

PrestaEdit:
| Questions         | Answers
| Description?      | Add the Currency resource to the Admin API
| BC breaks?        | no
| Deprecations?     | no
| Sponsor company   |

Adds the Admin API endpoints for the **Currency** domain:

- `POST /currencies` — `AddCurrencyCommand`
- `GET /currencies/{currencyId}` — `GetCurrencyForEditing`
- `PATCH /currencies/{currencyId}` — `EditCurrencyCommand`
- `DELETE /currencies/{currencyId}` — `DeleteCurrencyCommand`
- `PUT /currencies/{currencyId}/toggle-status` — `ToggleCurrencyStatusCommand`
- `PUT /currencies/bulk-toggle-status` — `BulkToggleCurrenciesStatusCommand`
- `DELETE /currencies/bulk-delete` — `BulkDeleteCurrenciesCommand`

Creating an official currency only requires `isoCode` + `exchangeRate` + `enabled`; the
localized names/symbols/transformations are filled from CLDR. The add/edit commands use
`isEnabled` + `localized*` field names, reconciled behind clean resource fields
(`enabled`, `names`, `symbols`, `transformations`); the query result's `associatedShopIds`
maps to `shopIds`.

Adds an integration test covering CRUD + toggle + bulk (on non-default currencies to
respect the default-currency constraints), and the `currency_read` / `currency_write` scopes.

### DIFF
--- a/src/ApiPlatform/Resources/Currency/BulkDeleteCurrencies.php
+++ b/src/ApiPlatform/Resources/Currency/BulkDeleteCurrencies.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\Currency;
+
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ApiResource;
+use PrestaShop\PrestaShop\Core\Domain\Currency\Command\BulkDeleteCurrenciesCommand;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSDelete;
+use Symfony\Component\Validator\Constraints as Assert;
+
+#[ApiResource(
+    operations: [
+        new CQRSDelete(
+            uriTemplate: '/currencies/bulk-delete',
+            CQRSCommand: BulkDeleteCurrenciesCommand::class,
+            scopes: [
+                'currency_write',
+            ],
+            allowEmptyBody: false,
+        ),
+    ],
+)]
+class BulkDeleteCurrencies
+{
+    /**
+     * @var int[]
+     */
+    #[ApiProperty(openapiContext: ['type' => 'array', 'items' => ['type' => 'integer'], 'example' => [1, 3]])]
+    #[Assert\NotBlank]
+    public array $currencyIds;
+}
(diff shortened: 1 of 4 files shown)


PrestaEdit:
Converting to draft. The integration tests surface a core/framework limitation: `EditableCurrency::isEnabled()` returns an **int** (unlike `EditableTax`/`EditableZone` which return a real `bool`), so the strict API Platform serializer rejects it for the `bool $enabled` resource property (`The type of the "enabled" attribute must be "bool", "integer" given` → 500 on GET). `ObjectNormalizer::DISABLE_TYPE_ENFORCEMENT` does not propagate to the CQRS query-result denormalization, and exposing `enabled` as an int would be inconsistent with every other resource (which return a bool). This needs either a small cast in the core `EditableCurrency` query result or a dedicated normalizer — out of scope for this PR. Holding until that's decided.

boo-code:
Hi @PrestaEdit — apologies, I opened #294 for the Currency resource before noticing yours was already in progress. Closing #294 in favour of this one.

Two things from #294 that may help here:
- **Core dependency:** the `enabled`/`isEnabled` field needs PrestaShop/PrestaShop#41884 — `GetCurrencyForEditingHandler` returns `$entity->active` as a string (`"1"`/`"0"`), so a strict bool deserialization fails without that cast (it's approved + milestoned for 9.2.0).
- I had an integration test (`CurrencyEndpointTest`, create/get/update/delete incl. localized names/symbols) — happy to adapt and contribute it here if useful.

PrestaEdit:
Thanks @boo-code, and no worries at all about #294 — much appreciated that you flagged it and closed in favour of this one. 🙏

On your two points:

- **`enabled` / PrestaShop/PrestaShop#41884:** nice catch on `GetCurrencyForEditingHandler` handing back `$entity->active` as a string. This PR already coerces it defensively through a `setEnabled(int|bool $enabled): self` setter (same approach as the `Title` resource's `setGender()`), so #235 isn't blocked and CI is green today. Once #41884 lands in 9.2.0 the setter becomes redundant — I'm happy to drop it as a follow-up, or keep it as a harmless safety net. Either way your core fix is the cleaner long-term solution. 👍
- **`CurrencyEndpointTest`:** this PR already ships an integration test covering create/get/update/delete + toggle + bulk, including localized names/symbols, so I think we're covered. If your version exercises edge cases mine doesn't, feel free to drop a diff and I'll happily fold them in.

Thanks again!

boo-code:
Thanks @PrestaEdit — agreed on both. The `setEnabled(int|bool)` setter is a sensible safety net; once #41884 lands in 9.2.0 it can be dropped, but no harm keeping it until then. 👍

Since you offered — two small cases my #294 test had that #235 doesn't seem to cover yet:
1. a validation test (empty `isoCode` → `422`), which also satisfies the `testInvalid…` convention in `CONTEXT.md`, and
2. a `404` on `GET /currencies/{missing}`.

Gist (standard `ApiTestCase` helpers; adapt the validation message to your `isoCode` constraint):

```php
public function testInvalidCurrency(): void
{
    $response = $this->createItem('/currencies', [
        'isoCode' => '',
        'exchangeRate' => 1.0,
        'enabled' => true,
    ], ['currency_write'], Response::HTTP_UNPROCESSABLE_ENTITY);

    $this->assertValidationErrors([
        ['propertyPath' => 'isoCode', 'message' => 'This value should not be blank.'],
    ], $response);
}

public function testGetMissingCurrencyReturnsNotFound(): void
{
    $this->getItem('/currencies/999999', ['currency_read'], Response::HTTP_NOT_FOUND);
}
```

Happy to open these as a tiny follow-up PR against your branch instead if that's easier. Your toggle + bulk coverage is already ahead of mine, so #235 is clearly the right home. Thanks!

PrestaEdit:
Folded both in — thanks @boo-code! 🙏

- `testInvalidCurrency`: empty `isoCode` → 422, asserting the single `isoCode` violation. The resource maps it via `#[Assert\NotBlank(groups: ['Create'])]`, so the default *"This value should not be blank."* message applies (same as `CountryEndpointTest`).
- 404 case: I named it `testGetNonExistentCurrency` to stay consistent with the existing `testGetNonExistentCountry` convention here, rather than the gist name — same coverage either way.

No need for the follow-up PR — it's in #235 now. Thanks again for the diff!

PrestaEdit:
Superseded by #423, which merges this endpoint with the rest of the domain, following the mutualisation asked for by the core team (one PR per domain rather than one per endpoint).

PR #237, opened 2026-06-15
Add Customer required-fields endpoint

PrestaEdit:
| Questions         | Answers
| Description?      | Add a Customer required-fields endpoint to the Admin API
| BC breaks?        | no
| Deprecations?     | no
| Sponsor company   |

Exposes `SetRequiredFieldsForCustomerCommand` through **`PUT /customers/required-fields`**,
which sets the list of customer fields that must be required (body: `requiredFields`). An
empty array clears them.

Adds an integration test (in its own class to isolate the global config change) and reuses
the `customer_write` scope.

### DIFF
--- a/src/ApiPlatform/Resources/Customer/CustomerRequiredFields.php
+++ b/src/ApiPlatform/Resources/Customer/CustomerRequiredFields.php
@@ -1,0 +1,49 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\Customer;
+
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ApiResource;
+use PrestaShop\PrestaShop\Core\Domain\Customer\Command\SetRequiredFieldsForCustomerCommand;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSUpdate;
+
+#[ApiResource(
+    operations: [
+        new CQRSUpdate(
+            uriTemplate: '/customers/required-fields',
+            output: false,
+            CQRSCommand: SetRequiredFieldsForCustomerCommand::class,
+            scopes: ['customer_write'],
+        ),
+    ],
+)]
+class CustomerRequiredFields
+{
+    /**
+     * The customer fields that should be required (e.g. "company", "siret").
+     *
+     * @var string[]
+     */
+    #[ApiProperty(openapiContext: ['type' => 'array', 'items' => ['type' => 'string'], 'example' => ['company', 'siret']])]
+    public array $requiredFields;
+}
--- a/src/ApiPlatform/Resources/Customer/CustomerRequiredFields.php
+++ b/src/ApiPlatform/Resources/Customer/CustomerRequiredFields.php
@@ -25,7 +25,10 @@ namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\Customer;
 use ApiPlatform\Metadata\ApiProperty;
 use ApiPlatform\Metadata\ApiResource;
 use PrestaShop\PrestaShop\Core\Domain\Customer\Command\SetRequiredFieldsForCustomerCommand;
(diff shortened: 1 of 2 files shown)


mattgoud (changes requested):
Thanks for the endpoint! The resource is well modelled (URI, `customer_write` scope and `output: false` are all correct). A few things need addressing before merge — one is a real blocker.

## 🔴 Blocker — missing `exceptionToStatus`

`CustomerRequiredFields` declares no `exceptionToStatus`, so a bad request surfaces as an uncaught **500**. Both exceptions this endpoint can raise extend `CustomerException` (**not** `CustomerConstraintException`):

- `InvalidCustomerRequiredFieldsException` — thrown in the `SetRequiredFieldsForCustomerCommand` constructor when a field name is not in `RequiredField::ALLOWED_REQUIRED_FIELDS` (the common client mistake);
- `CannotSetRequiredFieldsForCustomerException` — thrown by the handler if the DB write fails.

So mapping only `CustomerConstraintException => 422` would **not** catch the invalid-field case. Please map the two concrete exceptions:

```php
use PrestaShop\PrestaShop\Core\Domain\Customer\Exception\CannotSetRequiredFieldsForCustomerException;
use PrestaShop\PrestaShop\Core\Domain\Customer\Exception\InvalidCustomerRequiredFieldsException;
use Symfony\Component\HttpFoundation\Response;

#[ApiResource(
    operations: [ /* ... */ ],
    exceptionToStatus: [
        InvalidCustomerRequiredFieldsException::class => Response::HTTP_UNPROCESSABLE_ENTITY,
        CannotSetRequiredFieldsForCustomerException::class => Response::HTTP_UNPROCESSABLE_ENTITY,
    ],
)]
```

## 🟠 Uninitialised property

```php
public array $requiredFields;
```

If the request body omits `requiredFields`, the command's constructor (`__construct(array $requiredFields)`) never receives its argument and the request fails ungracefully. Defaulting to an empty array makes an omitted body behave as "clear all required fields", which matches the command's own empty-array semantics:

```php
public array $requiredFields = [];
```

## 🟠 Test coverage

The single test only sends an empty array (a no-op), so it exercises neither the command's validation nor the exception mapping above. Your own Address PR (#304) already has the right shape and works as a template here:

- a positive case that sets a real field (e.g. `['company']`) and asserts it is persisted in `required_field`;
- an invalid case (`['not_a_valid_field']`) asserting **422** — this is the test that proves the `exceptionToStatus` fix.

## ✅ No `CQRSCommandMapping` needed

Just confirming so it doesn't get added unnecessarily: the command constructor parameter is `requiredFields`, identical to the DTO property, so the default mapping is correct as-is.

---

Nice and focused otherwise. The `self::createApiClient(['customer_write'])` call is fine — it's consistent with #304 and the rest of the suite.

PrestaEdit:
Thanks for the thorough review — all three points addressed in the latest push:

- **Blocker**: added `exceptionToStatus` mapping both `InvalidCustomerRequiredFieldsException` and `CannotSetRequiredFieldsForCustomerException` to **422**. Confirmed both extend `CustomerException` (not `CustomerConstraintException`), so mapping the concrete exceptions is the only way to catch the invalid-field case.
- **Uninit property**: defaulted `public array $requiredFields = [];` — matches the command's empty-array-clears-all semantics for an omitted body.
- **Tests**: added a positive case that sets `['company']` and asserts it is persisted in `required_field` (with `object_name = 'Customer'`), plus an invalid case that sends `['not_a_valid_field']` and asserts **422** — the test that exercises the mapping fix.

mattgoud:
small correction on my earlier review, the `['company']` example i gave for the positive test was wrong on my part 🙏 `company`/`siret` are *address* required fields, for the customer required-fields command the only allowed values are `optin` and `newsletter` (`RequiredField::ALLOWED_REQUIRED_FIELDS`). so `testSetCustomerRequiredFields` should set e.g. `['newsletter']` and assert on that, and the DTO docblock/example should use `optin`/`newsletter` instead of `company`/`siret`.

the invalid-field test is fine as-is, it correctly returns 422 so your `exceptionToStatus` mapping works 👍 one thing worth a quick look: the positive test currently surfaces the invalid-field exception as a **500**, not 422, because it's thrown during deserialization (the command constructor, before the operation runs). once the positive test uses a valid field it won't throw at all, but might be worth confirming the 422 path is solid for that deser-time throw.

last thing, unrelated to your PR: the red CI on the 9.1.x/9.0.3 jobs is the OrderMessage `foreach()`-on-null bug, only fixed on `develop` so far. a backport is needed, i'm on it.

mattgoud (approved):
ok the test values are fixed and CI is green now 👍

also for the record: the red CI earlier was purely the `company` value in the positive test (my bad, i suggested `company` in my first review, it's not a valid customer required field). the OrderMessage warning on the 9.1.x/9.0.3 jobs was never what was failing the build. sorry for sending you down that path. LGTM.

mattgoud:
## 🧪 Functional QA — `PUT /customers/required-fields`

Tested manually via the Swagger UI on a local 9.2 install, with the module checked out on this PR branch. Also ran the integration test suite for the endpoint.

**Test matrix**

| Case | Request body | Auth | Expected | Result |
|------|--------------|------|----------|--------|
| Valid field | `{ "requiredFields": ["newsletter"] }` | `customer_write` | `204 No Content` + field persisted | ✅ |
| Invalid field | `{ "requiredFields": ["not_a_valid_field"] }` | `customer_write` | `422 Unprocessable Entity` | ✅ |
| Missing token | `{ "requiredFields": ["newsletter"] }` | none | `401 Unauthorized` | ✅ |

**Valid field → 204**

test-valid-field


**Invalid field → 422**

test-invalid-field-redacted


**Missing token → 401**

test-unauthorized

---

**Integration test**

```
CustomerRequiredFieldsEndpointTest .... OK (3 tests, 11 assertions)
```

QA passed ✅ — behavior matches the resource definition (204 on success, 422 on invalid field, scope-protected). LGTM.

mattgoud:
thanks @PrestaEdit

ps-jarvis:
PR merged, well done!

PR #243, opened 2026-06-16
Add Customer transform-guest-to-customer endpoint

PrestaEdit:
| Questions         | Answers
| Description?      | Add the transform-guest-to-customer endpoint
| BC breaks?        | no
| Deprecations?     | no
| Sponsor company   |

Adds **`PUT /customers/{customerId}/transform-to-customers`** — `TransformGuestToCustomerCommand`:
turns an existing guest customer into a registered customer (empty body).

- `404` if the customer does not exist (`CustomerNotFoundException`)
- `422` if the customer is not a guest (`CustomerTransformationException`)

Added as a **standalone resource class** (`TransformGuestToCustomer`) and a separate test, so it
doesn't collide with the in-progress Customer PRs that touch `Customer.php` / `CustomerEndpointTest.php`.

Adds an integration test (create a guest, transform it, then check a second transform is rejected)
and reuses the `customer_read` / `customer_write` scopes.

### DIFF
--- a/src/ApiPlatform/Resources/Customer/TransformGuestToCustomer.php
+++ b/src/ApiPlatform/Resources/Customer/TransformGuestToCustomer.php
@@ -1,0 +1,53 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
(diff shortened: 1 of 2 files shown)


PrestaEdit:
Superseded by #428, which merges this endpoint with the rest of the domain, following the mutualisation asked for by the core team (one PR per domain rather than one per endpoint).

PR #256, opened 2026-06-23
Add Product stock update endpoint

PrestaEdit:
| Questions         | Answers
| Description?      | Add the Product stock update endpoint
| BC breaks?        | no
| Deprecations?     | no
| Sponsor company   |

Adds **`PUT /products/{productId}/stocks`** — `UpdateProductStockAvailableCommand`: adjusts a
product's available stock via `deltaQuantity` (positive to add, negative to remove), and optionally
its out-of-stock behaviour (`outOfStockType`) and stock `location`. The shop constraint is taken
from the request context. `404` when the product does not exist.

### ⚠️ Depends on a core fix — kept as a draft

A stock quantity change records a `StockMvt` attributed to the current employee. In the API context
there is no logged-in employee, so `StockManager` called `StockMvt::setIdEmployee(null)` →
`TypeError`. The core PR **PrestaShop/PrestaShop#41803** guards that (keeps `id_employee = 0` when the
context employee has no id). This PR stays a draft and red on the released-core matrices until that
fix ships; it goes green on `develop` once #41803 is merged.

The integration test picks a simple fixture product (no combinations), applies a `+5` delta and
verifies the resulting quantity via `StockAvailable`. Reuses the `product_write` scope.

### DIFF
--- a/src/ApiPlatform/Resources/Product/ProductStock.php
+++ b/src/ApiPlatform/Resources/Product/ProductStock.php
@@ -1,0 +1,61 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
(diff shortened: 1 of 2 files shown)


jolelievre:

Note: the URI becomes PUT /products/{productId}/stock (singular — a product has one stock) in #410, and the operation returns the updated stock state. The red CI of this PR was actually caused by a core bug: any stock update through the Admin API hits a StockMvt::setIdEmployee(null) TypeError — the fix is PrestaShop/PrestaShop#41803, which #410 depends on.

PR #268, opened 2026-06-23
Add Product suppliers endpoints (associate, default, remove all)

PrestaEdit:
| Questions         | Answers
| Description?      | Add the Product suppliers endpoints (associate, default, remove all)
| BC breaks?        | no
| Deprecations?     | no
| Sponsor company   |

Adds the Admin API endpoints to manage a product's supplier associations:

- `PUT /products/{productId}/suppliers` — `SetSuppliersCommand` (associate suppliers by id)
- `DELETE /products/{productId}/suppliers` — `RemoveAllAssociatedProductSuppliersCommand`
- `PUT /products/{productId}/default-suppliers` — `SetProductDefaultSupplierCommand`

These manage the associations only (per-supplier reference/price content is handled by a separate
command and can follow up). All return `204`.

The integration test associates a fixture supplier, sets it as the default (verified via
`product.id_supplier`) and removes all (verified via the `product_supplier` table). Reuses the
`product_write` scope.

### DIFF
--- a/src/ApiPlatform/Resources/Product/ProductSuppliers.php
+++ b/src/ApiPlatform/Resources/Product/ProductSuppliers.php
@@ -1,0 +1,82 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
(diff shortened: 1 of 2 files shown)


jolelievre:

Note: in #410 the three operations live in the consolidated ProductSuppliers resource together with the GET and the details PATCH, every write returns the updated supplier options, and the default supplier URI is singularized to /products/{productId}/default-supplier.

PR #284, opened 2026-06-26
Skip OrderMessage messages-only test until core #41790 ships

PrestaEdit:
| Questions         | Answers
| Description?      | Skip the OrderMessage messages-only test until core #41790 ships
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Sponsor company   |

`OrderMessageEndpointTest::testEditOrderMessageMessagesOnly` (added in #216) does a messages-only
`PATCH`. That reaches the core `EditOrderMessageHandler::assertNameIsNotAlreadyUsed()`, which iterates
over the (null) localized name and emits a `foreach()` warning. On the warning-as-error integration
matrices this fails the build — for **every** PR branched off `dev`, not just OrderMessage ones.

This skips that single test until the core guard **PrestaShop/PrestaShop#41790** is released (it stays
covered by the create / get / names-only / delete cases). Once #41790 ships, the skip can be removed
and the assertion restored.

### DIFF
--- a/tests/Integration/ApiPlatform/OrderMessageEndpointTest.php
+++ b/tests/Integration/ApiPlatform/OrderMessageEndpointTest.php
@@ -183,6 +183,11 @@ class OrderMessageEndpointTest extends ApiTestCase
 
     public function testEditOrderMessageMessagesOnly(): void
     {
+        // A messages-only edit reaches EditOrderMessageHandler::assertNameIsNotAlreadyUsed() which
+        // iterates over the (null) localized name and emits a foreach() warning, failing the build on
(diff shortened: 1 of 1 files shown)


PrestaEdit:
Closing: `dev` already skips this test, and with a better guard than the one proposed here.

This PR skips `testEditOrderMessageMessagesOnly` unconditionally. `dev` uses `markTestSkippedByMinVersion('9.2.0')`, so the test is skipped only on the cores that lack the guard from PrestaShop/PrestaShop#41790 and runs again from 9.2.0 on — which is what we want now that the core fix has shipped.

PR #300, opened 2026-06-30
Add categories tree read endpoint

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?       | Adds a read endpoint exposing the category tree: `GET /categories/trees`. It returns the localized category tree for the current shop and language as a nested collection — each node exposes `categoryId`, `enabled`, `name`, `displayName` and its `children` (same shape, recursively). This is the `GetCategoriesTree` query used by the BO category pickers, previously unexposed. Scope `category_read`.
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | Call `GET /categories/trees` with a client granted the `category_read` scope and check the payload is a nested tree where each node exposes `categoryId`, `enabled`, `name`, `displayName` and `children`. Covered by `CategoryTreeEndpointTest`.

### DIFF
--- a/src/ApiPlatform/Resources/Category/Category.php
+++ b/src/ApiPlatform/Resources/Category/Category.php
@@ -45,6 +45,7 @@ use Symfony\Component\Validator\Constraints as Assert;
     operations: [
         new CQRSGet(
             uriTemplate: '/categories/{categoryId}',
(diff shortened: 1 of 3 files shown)


PrestaEdit:

PR #314, opened 2026-07-06
Add cart send-to-customer endpoint

### LINKED ISSUE #41969: Admin API - Add missing endpoints for the "Cart" domain
Endpoint:
- SendCartToCustomerCommand

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?       | Fills the `SendCartToCustomerCommand` gap: `PUT /carts/{cartId}/emails` (scope `cart_write`) to send the cart to its customer by email.
| BC breaks?        | no
| Deprecations?     | no
| Related ticket?     | Related to PrestaShop/PrestaShop#39630
| Fixed ticket?     | fixed https://github.com/PrestaShop/PrestaShop/issues/41969
| How to test?      | `PUT /carts/{id}/emails` (client granted `cart_write`) emails the cart to the customer. Covered by `CartEmailEndpointTest` (which disables real mail sending so the assertion does not depend on an SMTP server).
| Possible impacts? | Sends an email to the cart customer.

### DIFF
--- a/src/ApiPlatform/Resources/Cart/CartEmail.php
+++ b/src/ApiPlatform/Resources/Cart/CartEmail.php
@@ -18,6 +18,8 @@
  * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
  */
 
+declare(strict_types=1);
+
 namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\Cart;
 
 use ApiPlatform\Metadata\ApiProperty;
--- a/src/ApiPlatform/Resources/Cart/CartEmail.php
+++ b/src/ApiPlatform/Resources/Cart/CartEmail.php
@@ -1,0 +1,54 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\Cart;
+
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ApiResource;
+use PrestaShop\PrestaShop\Core\Domain\Cart\Command\SendCartToCustomerCommand;
+use PrestaShop\PrestaShop\Core\Domain\Cart\Exception\CartException;
+use PrestaShop\PrestaShop\Core\Domain\Cart\Exception\CartNotFoundException;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSUpdate;
+use Symfony\Component\HttpFoundation\Response;
+
+#[ApiResource(
+    operations: [
+        new CQRSUpdate(
+            uriTemplate: '/carts/{cartId}/emails',
+            requirements: ['cartId' => '\d+'],
+            read: false,
+            output: false,
+            allowEmptyBody: true,
+            CQRSCommand: SendCartToCustomerCommand::class,
+            scopes: [
+                'cart_write',
+            ],
+        ),
+    ],
+    exceptionToStatus: [
+        CartNotFoundException::class => Response::HTTP_NOT_FOUND,
+        CartException::class => Response::HTTP_UNPROCESSABLE_ENTITY,
+    ],
+)]
+class CartEmail
+{
+    #[ApiProperty(identifier: true)]
+    public int $cartId;
+}
--- a/tests/Integration/ApiPlatform/CartEmailEndpointTest.php
+++ b/tests/Integration/ApiPlatform/CartEmailEndpointTest.php
@@ -63,6 +63,21 @@ class CartEmailEndpointTest extends ApiTestCase
         );
     }
 
+    public function testSendUnknownCartReturnsNotFound(): void
+    {
+        $unknownCartId = 1 + (int) \Db::getInstance()->getValue(
+            'SELECT MAX(`id_cart`) FROM `' . _DB_PREFIX_ . 'cart`'
+        );
+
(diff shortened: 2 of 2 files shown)


mattgoud (changes requested):
Thanks for this. The endpoint is correct and functional and follows the module conventions well:

- `SendCartToCustomerCommand::__construct($cartId)` maps cleanly from the `cartId` identifier, no `CQRSCommandMapping` needed, and the green integration test confirms the wiring.
- The exception mapping is correctly ordered: `CartNotFoundException => 404` before `CartException => 422` (required since `CartNotFoundException extends CartException`).
- `read: false` + `output: false` + `allowEmptyBody: true` is the right combo for a no-input/no-output action returning `204`, and disabling `PS_MAIL_METHOD` in the test avoids an SMTP dependency.

A few changes I would like before merging:

1. **Add `declare(strict_types=1);` to `CartEmail.php`.** The test file already has it, and it is the project standard for every PHP file.
2. **Cover the not-found case in the test.** A `PUT /carts/{nonExistentId}/emails` asserting `404` would exercise the `CartNotFoundException => 404` mapping, which is the most subtle part of the resource and is currently untested. This is the main one for me.
3. **Use a non-empty cart in `createCustomerCart()`** (add a product) so the test is more representative of a real send-to-customer flow.

Happy to re-review once these are in.

PrestaEdit:
Thanks for the thorough review @mattgoud! Addressed all three nits in dc08b06:
- Added `declare(strict_types=1);` to the `CartEmail` resource.
- Added `testSendUnknownCartReturnsNotFound` — PUTs to a non-existent `cartId` and asserts `404`, exercising the `CartNotFoundException => 404` mapping.
- The test cart now holds a product, so it's a more representative send-to-customer case.

mattgoud (approved):
Perfect, all three addressed in dc08b06 and CI is green across the matrix:

- `declare(strict_types=1);` added to `CartEmail.php`.
- `testSendUnknownCartReturnsNotFound` uses `MAX(id_cart)+1` for a guaranteed non-existent id and asserts `404`, so the `CartNotFoundException => 404` mapping is now exercised end-to-end.
- `createCustomerCart()` now adds a simple, active, in-stock product (nicely filtered out of products with combinations), which is more representative of a real send-to-customer flow.

Thanks for the quick turnaround. LGTM 👍

mattgoud:
Following up on the AI pre-review checklist — not every box is ticked, but they are not all equal, so here is my triage so you know what is worth acting on vs noise:

**Worth fixing (the one I would actually like addressed):**
- **`DatabaseDump::restoreTables()` is missing.** The test inserts into `cart` (and `cart_product` via `updateQty`) but never restores them, so the rows leak into other test classes — including any that read `MAX(id_cart)`, which your own `testSendUnknownCartReturnsNotFound` relies on. 29 endpoint tests in the module already do this. Adding `DatabaseDump::restoreTables(['cart', 'cart_product']);` in both `setUpBeforeClass` and `tearDownAfterClass` would align it. I missed this in my first pass — good catch by the bot.

**Design calls, your/maintainers' decision (not blocking):**
- **URI naming `emails` vs a verb suffix**, and **PUT vs POST**. The module leans towards verb suffixes for action sub-resources (`/modules/{name}/reset`, `/zones/{id}/toggle-status`), so `/carts/{cartId}/send-to-customer` would read more clearly, and POST would be more honest about the non-idempotent side effect. That said, `Zone` already uses `CQRSUpdate + output:false` for an action, so the current choice is defensible. Up to you.
- **`testInvalid*` / 422**: only reachable if the endpoint can surface a `CartException` (e.g. a cart with no valid customer). Either add one case or drop a one-line comment saying why it is N/A here.

**Bot noise — safe to ignore:**
- "Remove the explicit `createApiClient`" — this is actually the dominant convention (36 test classes call it in `setUpBeforeClass`), so keep it.
- `validationContext` / `@depends` — N/A for a no-body single-action endpoint, as the bot itself notes.

My approval stands; only the DB cleanup is something I would genuinely like in before merge.

mattgoud (changes requested):
Switching my approval back to request-changes to gate the merge on the one real item from the AI pre-review (my earlier comment stands for the full triage).

**Blocking:**
- **Add `DatabaseDump::restoreTables(['cart', 'cart_product']);`** in both `setUpBeforeClass` and `tearDownAfterClass`. The test inserts into `cart` / `cart_product` and currently never restores them, leaking rows into other test classes — including any reading `MAX(id_cart)`, which `testSendUnknownCartReturnsNotFound` itself depends on. 29 endpoint tests in the module already follow this pattern.

Everything else (URI naming, PUT vs POST, `testInvalid*`) is a non-blocking design call, and the bot's "remove `createApiClient`" suggestion is wrong (it is the dominant convention). Happy to re-approve as soon as the DB cleanup is in.

PrestaEdit:
Thanks @mattgoud, appreciate the triage.

- **DB cleanup** — done in d8e64a8: `DatabaseDump::restoreTables(['cart', 'cart_product']);` now runs in both `setUpBeforeClass` and `tearDownAfterClass`, so the inserted rows no longer leak (and the `MAX(id_cart)` lookup in the 404 test stays reliable).
- **422 case** — added a one-line note in the test (2c71cb9) explaining why it's N/A: the only path to a generic `CartException` is a mail-send failure or a cart whose customer no longer exists, neither of which can be reliably set up through this endpoint in an integration test.
- **URI naming / PUT vs POST** — happy to switch to `/carts/{cartId}/send-to-customer` + POST if you'd prefer, but since `Zone` already uses the same `CQRSUpdate + output:false` action pattern I've left it as-is. Your call.

mattgoud (approved):
Re-approving. Both points are in and CI is green across the full matrix:

- **DB cleanup (was blocking):** `DatabaseDump::restoreTables(['cart', 'cart_product'])` added to `setUpBeforeClass` and `tearDownAfterClass` (d8e64a8). No more row leakage.
- **422 mapping (non-blocking):** documented why the `CartException => 422` path is not reachable from an integration test (2c71cb9).

Thanks again for the quick turnarounds. LGTM 👍

tleon (approved):

SiraDIOP (commented):
Scenario 1 — Successful email send

Sent PUT /admin-api/carts/6/emails (client with cart_write scope, empty body) against an existing cart (id_cart=6) with a customer and a product attached.

Response: 204 No Content.

Verified via MailDev (localhost:1080) that an email was actually delivered to the cart's customer, timestamped to match the request. The email uses the "Process the payment of your order" template with a recover_cart link, which is the expected behavior for SendCartToCustomerCommand.

✅ Working as expected.

Image

Image

Scenario 2 — Non-existent cart (404)

Sent PUT /admin-api/carts/99999/emails against a non-existent cart ID.

Response: 404 Not Found (HTTP status), with a CartNotFoundException correctly thrown and a clear detail message ("Cart with id "99999" was not found"). Note: the JSON body's "status" field shows 500, but this is a known/expected serialization detail unrelated to this PR.

✅ Working as expected.

Image

Thanks

SiraDIOP (approved):

ps-jarvis:
PR merged, well done!

PR #316, opened 2026-07-06
Add order resend-email endpoint

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?       | Fills the `ResendOrderEmailCommand` gap: `PUT /orders/{orderId}/emails` (scope `order_write`) to resend the customer email of an order status history entry (`orderStatusId`, `orderHistoryId`).
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | `PUT /orders/{id}/emails` with `{ "orderStatusId": x, "orderHistoryId": y }` (client granted `order_write`) resends the status email. Covered by `OrderEmailEndpointTest` (which disables real mail sending so the assertion does not depend on an SMTP server).
| Possible impacts? | Resends an order status email to the customer.

### DIFF
--- a/src/ApiPlatform/Resources/Order/OrderEmail.php
+++ b/src/ApiPlatform/Resources/Order/OrderEmail.php
@@ -1,0 +1,59 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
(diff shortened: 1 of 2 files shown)


mattgoud (changes requested):
The resource looks good, but the test only covers `getProtectedEndpoints()` + the happy path. The module testing conventions require a `testInvalid*` (422) case as well — e.g. a bad `orderStatusId` / `orderHistoryId` should surface as `HTTP_UNPROCESSABLE_ENTITY`, asserted via `assertValidationErrors(...)`.

PrestaEdit:

Rather than only asserting the handler-thrown `OrderException => 422`, I moved the check up to the resource so the response body matches the standard validation payload:

- Added `#[Assert\Positive]` on `$orderStatusId` and `$orderHistoryId` on the `OrderEmail` resource — `0` (or a negative value) is now rejected at the Symfony validator layer, before the command hits the handler.
- Added `testResendOrderEmailWithInvalidPayload` — PUTs with `orderStatusId: 0, orderHistoryId: 0`, asserts `HTTP_UNPROCESSABLE_ENTITY`, and uses `assertValidationErrors(...)` on the returned `{propertyPath, message, code}` structure.

Ready for another look.

mattgoud (approved):
Thanks, the 422 negative test is in and CI is green. LGTM.

PR #317, opened 2026-07-06
Add order update-status endpoint

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?       | Fills the `UpdateOrderStatusCommand` gap: `PUT /orders/{orderId}/status` (scope `order_write`) to change an order's current status (`newOrderStatusId`).
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | `PUT /orders/{id}/status` with `{ "newOrderStatusId": x }` (client granted `order_write`) changes the order status. Covered by `OrderStatusEndpointTest` (which disables real mail sending so the assertion does not depend on an SMTP server).
| Possible impacts? | Changes the order status (and sends the associated status email).

### DIFF
--- a/src/ApiPlatform/Resources/Order/OrderStatus.php
+++ b/src/ApiPlatform/Resources/Order/OrderStatus.php
@@ -23,7 +23,6 @@ namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\Order;
 use ApiPlatform\Metadata\ApiProperty;
 use ApiPlatform\Metadata\ApiResource;
 use PrestaShop\PrestaShop\Core\Domain\Order\Command\UpdateOrderStatusCommand;
-use PrestaShop\PrestaShop\Core\Domain\Order\Exception\ChangeOrderStatusException;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\OrderException;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\OrderNotFoundException;
 use PrestaShopBundle\ApiPlatform\Metadata\CQRSUpdate;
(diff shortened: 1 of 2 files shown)


mattgoud (changes requested):
The resource looks good, but the test only covers `getProtectedEndpoints()` + the happy path. Please add a `testInvalid*` (422) case per the module testing conventions — e.g. an invalid `newOrderStatusId` (or an illegal status transition) returning `HTTP_UNPROCESSABLE_ENTITY`.

PrestaEdit:

Same shape as [#316](https://github.com/PrestaShop/ps_apiresources/pull/316): moved the check up to the resource so the response body matches the standard validation payload rather than the raw handler-thrown `OrderException`.

- Added `#[Assert\Positive]` on `OrderStatus::$newOrderStatusId` — `0` (or a negative value) is now rejected at the Symfony validator layer, before the command hits the handler.
- Added `testUpdateOrderStatusWithInvalidPayload` — PUTs with `newOrderStatusId: 0`, asserts `HTTP_UNPROCESSABLE_ENTITY`, and uses `assertValidationErrors(...)` on the `{propertyPath, message, code}` structure.

Ready for another look.

mattgoud:
Tiny cosmetic nit (non-blocking): the `ChangeOrderStatusException => 422` entry in `exceptionToStatus` is redundant. `ChangeOrderStatusException extends OrderException`, and you already map `OrderException => 422`, so the specific entry can never change the outcome. The `OrderException` catch-all is correct to keep here (the handler throws `new OrderException(...)` directly), so you can just drop the `ChangeOrderStatusException` line to keep the map tidy.

mattgoud (approved):
Thanks, the redundant ChangeOrderStatusException entry is gone and the 422 test is in. LGTM.

PR #318, opened 2026-07-06
Add cart process-order-email endpoint

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?       | Fills the `SendProcessOrderEmailCommand` gap: `PUT /carts/{cartId}/process-order-emails` (scope `order_write`) to send the process-order email to the cart's customer.
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | `PUT /carts/{id}/process-order-emails` (client granted `order_write`) emails the customer. Covered by `CartProcessOrderEmailEndpointTest` (which disables real mail sending so the assertion does not depend on an SMTP server).
| Possible impacts? | Sends the process-order email to the cart customer.

### DIFF
--- a/src/ApiPlatform/Resources/Cart/CartProcessOrderEmail.php
+++ b/src/ApiPlatform/Resources/Cart/CartProcessOrderEmail.php
@@ -22,6 +22,7 @@ namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\Cart;
 
 use ApiPlatform\Metadata\ApiProperty;
 use ApiPlatform\Metadata\ApiResource;
+use PrestaShop\PrestaShop\Core\Domain\Cart\Exception\CartConstraintException;
 use PrestaShop\PrestaShop\Core\Domain\Cart\Exception\CartNotFoundException;
 use PrestaShop\PrestaShop\Core\Domain\Customer\Exception\CustomerNotFoundException;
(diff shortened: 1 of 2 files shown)


mattgoud (changes requested):
`CartConstraintException` is missing from `exceptionToStatus`. The command builds `new CartId($cartId)`, which throws `CartConstraintException` (extends `CartException`, not `OrderException`) for an invalid id such as `0` — and `0` is reachable since the `\d+` requirement matches it. It currently falls through to a 500 instead of 422. Please add `CartConstraintException::class => Response::HTTP_UNPROCESSABLE_ENTITY`.

Also missing a `testInvalid*` (422) test per the module testing conventions.

PrestaEdit:

- Added `CartConstraintException::class => Response::HTTP_UNPROCESSABLE_ENTITY` to `exceptionToStatus` on the resource, exactly as flagged. Since `\d+` in the URI matches `0`, `new CartId(0)` used to bubble up as a 500; it now returns 422 cleanly.
- Added `testSendProcessOrderEmailWithInvalidCartIdReturns422` — PUTs `/carts/0/process-order-emails` and asserts 422.

Ready for another look.

mattgoud (approved):
Thanks, CartConstraintException is now mapped to 422 and the negative test is in. LGTM.

PrestaEdit:
Superseded by #424, which merges this endpoint with the rest of the domain, following the mutualisation asked for by the core team (one PR per domain rather than one per endpoint).

PR #319, opened 2026-07-06
Add order bulk-update-status endpoint

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?       | Fills the `BulkChangeOrderStatusCommand` gap: `PUT /orders/bulk-update-status` (scope `order_write`) to change the status of several orders at once (`orderIds`, `newOrderStatusId`).
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | `PUT /orders/bulk-update-status` with `{ "orderIds": [id1, id2], "newOrderStatusId": x }` (client granted `order_write`) changes all the given orders' status. Covered by `BulkOrderStatusEndpointTest` (which disables real mail sending so the assertion does not depend on an SMTP server).
| Possible impacts? | Changes the status (and sends the status emails) of the given orders.

### DIFF
--- a/src/ApiPlatform/Resources/Order/BulkOrderStatus.php
+++ b/src/ApiPlatform/Resources/Order/BulkOrderStatus.php
@@ -1,0 +1,54 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\Order;
+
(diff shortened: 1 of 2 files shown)


mattgoud (changes requested):
The resource looks good, but the test only covers `getProtectedEndpoints()` + the happy path. Please add a `testInvalid*` (422) case per the module testing conventions — e.g. empty `orderIds` or an invalid `newOrderStatusId` returning `HTTP_UNPROCESSABLE_ENTITY`.

PrestaEdit:

Same shape as [#316](https://github.com/PrestaShop/ps_apiresources/pull/316) / [#317](https://github.com/PrestaShop/ps_apiresources/pull/317): moved the checks up to the resource so the response body matches the standard validation payload.

- Added `#[Assert\NotBlank]` + `#[Assert\All([new Assert\Positive()])]` on `$orderIds` — empty array or any non-positive id is rejected at the Symfony validator layer.
- Added `#[Assert\Positive]` on `$newOrderStatusId`.
- Added `testBulkChangeOrderStatusWithEmptyOrderIds` and `testBulkChangeOrderStatusWithInvalidStatusId`, both using `assertValidationErrors(...)` on the `{propertyPath, message, code}` structure.

Ready for another look.

mattgoud:
Two small conformance points (surfaced by the AI pre-review, verified against the code):

1. **`$orderIds` is missing its `#[ApiProperty(openapiContext: ...)]`.** CONTEXT.md requires array properties to document their item shape for the OpenAPI schema, and the existing bulk resources all do it. Suggested:
   ```php
   #[ApiProperty(openapiContext: ['type' => 'array', 'items' => ['type' => 'integer']])]
   #[Assert\NotBlank]
   #[Assert\All([new Assert\Positive()])]
   public array $orderIds;
   ```
2. **`declare(strict_types=1);` is missing on `BulkOrderStatus.php`.** The project rule is every PHP file declares it (other module resources do). Please add it after the license header.

mattgoud (approved):
Re-reviewed 👍 both conformance points are in (`openapiContext` on `$orderIds` + `declare(strict_types=1)`), and both 422 cases are covered (empty `orderIds` + non-positive `newOrderStatusId`). Happy-path restores state cleanly. LGTM

PR #320, opened 2026-07-06
Add order partial-refund endpoint

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?       | Fills the `IssuePartialRefundCommand` gap: `POST /orders/{orderId}/partial-refunds` (scope `order_write`) to issue a partial refund on an order (`orderDetailRefunds` keyed by order detail id with `quantity`/`amount`, `shippingCostRefundAmount`, `restockRefundedProducts`, `generateCreditSlip`, `generateVoucher`, `voucherRefundType`, `voucherRefundAmount`).
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | `POST /orders/{id}/partial-refunds` with a refund payload (client granted `order_write`) issues the refund. Covered by `OrderPartialRefundEndpointTest` (which disables real mail sending and uses a paid+delivered order with `restockRefundedProducts=false`).
| Possible impacts? | Records a partial refund on the order (and optionally a credit slip / voucher).

### DIFF
--- a/src/ApiPlatform/Resources/Order/OrderPartialRefund.php
+++ b/src/ApiPlatform/Resources/Order/OrderPartialRefund.php
@@ -1,0 +1,72 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\Order;
+
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ApiResource;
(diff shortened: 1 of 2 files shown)


mattgoud (changes requested):
The complex array property `orderDetailRefunds` is missing the `#[ApiProperty(openapiContext: ['type' => 'array', 'items' => [...]])]` annotation the module conventions require for complex array structures, so the OpenAPI schema doesn't document its shape. (For reference, the expected body shape is `{orderDetailId: {quantity, amount}}`, which your test already uses correctly.)

Also missing a `testInvalid*` (422) test.

PrestaEdit:

- Added `#[ApiProperty(openapiContext: [...])]` on `$orderDetailRefunds` documenting the `{orderDetailId: {quantity, amount}}` shape (`type: object` + `additionalProperties` with the nested `quantity`/`amount` schema and an example).
- Added `#[Assert\NotBlank]` on the same property so an empty refund set is rejected at the validator layer with the standard 422 payload.
- Added `testIssuePartialRefundWithEmptyOrderDetailRefunds` — PUTs with `orderDetailRefunds: []`, asserts 422 via `assertValidationErrors(...)`.

Ready for another look.

mattgoud:
One design point worth aligning before merge (raised by the AI pre-review): this issues a partial refund via `CQRSUpdate` (PUT), but the action is **not idempotent** — two identical calls create two credit slips / vouchers. PUT implies a safe-to-repeat replace, which doesn't hold here.

`#324` (add-discount) already switched from PUT to `CQRSCreate` (POST) for exactly this reason; the same applies to a refund. Recommend `CQRSCreate` (POST) here too (with `output: false`, as you're doing).

(For the record, the pre-review also suggested `DecimalNumber` for the amount fields — that one can be ignored: `IssuePartialRefundCommand::__construct` takes `string` for `shippingCostRefundAmount` / `voucherRefundAmount`, so `string` on the DTO is correct.)

PrestaEdit:
Good catch, agreed — switched to `CQRSCreate` (POST) in the resource and updated the endpoint test / PR description to match. Pushed as [commit on the branch]. Thanks!

mattgoud (approved):
Thanks, switched to CQRSCreate (POST) so the action is no longer a non-idempotent PUT. LGTM.

PR #322, opened 2026-07-06
Fix AI pre-review being cancelled by unrelated label events

### LINKED ISSUE #321: AI Pre-Review: unrelated label events cancel an in-progress review (concurrency group too broad)
The `Claude Pre-Review` workflow (`.github/workflows/ai-prereview.yml`) can be silently cancelled mid-review, leaving the `Need AI review` label in place and never swapping it to `AI reviewed`.

### Root cause

The workflow triggers on **any** `labeled` event and uses a concurrency group keyed only on the PR number:

```yaml
on:
  pull_request_target:
    types: [labeled]

concurrency:
  group: ai-prereview-${{ github.event.pull_request.number }}
  cancel-in-progress: true
```

The actual work is gated later by `if: github.event.label.name == 'Need AI review'`. But `concurrency` is evaluated **before** that job-level `if`. So when *another* label is added to the same PR while a review is running, its (otherwise no-op) run joins the same concurrency group and, with `cancel-in-progress: true`, **cancels the running review**. The cancelling run then skips all jobs because its label is not `Need AI review`, so it neither reviews nor swaps the label.

### Observed occurrence (PR #314)

| Time (UTC) | Event | Run | Result |
|-----------|-------|-----|--------|
| 13:46:34 | `Need AI review` added | 28796300493 | guard passed, `Claude AI review` **started** then **cancelled** |
| 13:49:09 | `Waiting for author` added by `ps-jarvis` | 28796461235 | all jobs **skipped** (label != `Need AI review`) |

The `Waiting for author` label (added automatically by the status bot) landed ~2.5 min into the review and cancelled it via the shared concurrency group.

### Impact

- Intermittent: only happens when a second `labeled` event hits the PR during the review window (~1-3 min).
- The label is never swapped, so the PR looks like the AI review is still pending when it silently died.

### Proposed fix

Include the label name in the concurrency group so events for different labels no longer collide, while genuine re-triggers of `Need AI review` still cancel the stale run (the intended behaviour):

```yaml
concurrency:
  group: ai-prereview-${{ github.event.pull_request.number }}-${{ github.event.label.name }}
  cancel-in-progress: true
```

mattgoud:
| Questions         | Answers
| Branch?           | dev
| Description?      | The `Claude Pre-Review` workflow fires on any `labeled` event but its concurrency group was keyed only on the PR number with `cancel-in-progress: true`. A label added for another reason (e.g. `Waiting for author`, auto-added by the status bot) therefore cancelled an in-progress `Need AI review` run. The cancelling run then skipped all jobs (its label is not `Need AI review`), so the review neither completed nor swapped the label. This scopes the concurrency group to the label name so events for different labels no longer collide.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #321
| How to test?      | Add `Need AI review` to a PR, then add another label (or let the status bot add `Waiting for author`) while the review is running. Before: the review run is cancelled and the label is never swapped to `AI reviewed`. After: the second label gets its own concurrency group and the review completes and swaps the label. Genuine re-triggers of `Need AI review` still cancel the previous run as intended.
| Possible impacts? | CI only (the AI pre-review workflow). No runtime/module code changed.

Observed on PR #314: the `Need AI review` run started, passed the guard and began the Claude review, then was cancelled ~2.5 min in by a `Waiting for author` label added by `ps-jarvis`.

### DIFF
--- a/.github/workflows/ai-prereview.yml
+++ b/.github/workflows/ai-prereview.yml
@@ -5,7 +5,10 @@ on:
     types: [labeled]
 
 concurrency:
-  group: ai-prereview-${{ github.event.pull_request.number }}
+  # Scope the group to the label too: the workflow fires on any `labeled` event but
+  # only does work for `Need AI review`. Keying on the PR number alone let an unrelated
+  # label (e.g. `Waiting for author`) cancel an in-progress review via cancel-in-progress.
+  group: ai-prereview-${{ github.event.pull_request.number }}-${{ github.event.label.name }}
   cancel-in-progress: true
 
 jobs:


cnavarro-prestashop (approved):

ps-jarvis:
PR merged, well done!

PR #324, opened 2026-07-06
Add order discount endpoint

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?       | Fills the `AddCartRuleToOrderCommand` gap: `PUT /orders/{orderId}/discounts` (scope `order_write`) to add a discount to an order (`cartRuleName`, `cartRuleType` = percent/amount/free_shipping, `value`, optional `orderInvoiceId`).
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | `PUT /orders/{id}/discounts` with `{ "cartRuleName": "...", "cartRuleType": "amount", "value": "1" }` (client granted `order_write`) adds the discount. Covered by `OrderDiscountEndpointTest`.
| Possible impacts? | Adds a cart rule (discount) to the order and recalculates its totals.

### DIFF
--- a/src/ApiPlatform/Resources/Order/OrderDiscount.php
+++ b/src/ApiPlatform/Resources/Order/OrderDiscount.php
@@ -1,0 +1,66 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\Order;
+
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ApiResource;
+use PrestaShop\PrestaShop\Core\Domain\CartRule\Exception\InvalidCartRuleDiscountValueException;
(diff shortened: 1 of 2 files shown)


mattgoud (changes requested):
This wraps an `Add…Command` but exposes it as `CQRSUpdate` (PUT) on a plural collection URI `/orders/{orderId}/discounts`. The operation is non-idempotent (calling it twice adds two discounts), so per the module conventions it should be `CQRSCreate` (POST).

Also missing a `testInvalid*` (422) test — empty `cartRuleName` or a null `value` for an amount/percent discount both throw `OrderConstraintException` (422) and would make good negative cases.

PrestaEdit:

- Switched `CQRSUpdate` (PUT) → `CQRSCreate` (POST) on the same URI, so the operation now matches its non-idempotent semantics (calling twice adds two discounts). Kept `output: false` since there's no CQRSQuery for a single order-cart-rule to return.
- Added `#[Assert\NotBlank]` on `$cartRuleName` and `#[Assert\Choice(choices: ['percent', 'amount', 'free_shipping'])]` on `$cartRuleType` so malformed payloads are rejected at the validator layer with the standard 422 payload.
- Test: switched `testAddCartRuleToOrder` to `createItem` (POST) with `HTTP_CREATED`; updated `getProtectedEndpoints` to POST; added `testAddCartRuleWithEmptyName` asserting 422 via `assertValidationErrors` on `cartRuleName`.

Ready for another look.

mattgoud:
Thanks for switching to `CQRSCreate` (POST). CI is red now though: the happy-path `testAddCartRuleToOrder` asserts `HTTP_CREATED` (201), but with `output: false` the endpoint returns `204 No Content`, so the assertion fails on all 8 integration matrix jobs. Either assert `HTTP_NO_CONTENT` (204), or, if you want the created state echoed back, drop `output: false` and wire a `CQRSQuery`.

mattgoud:
Two small robustness/convention points (from the AI pre-review, verified):

1. **`$cartRuleType` is required but only has `#[Assert\Choice(...)]`, no `#[Assert\NotBlank]`.** Since the property is a non-nullable `string` with no default, omitting it from the request risks an uninitialized-typed-property error (500) instead of a clean 422. Adding `#[Assert\NotBlank]` gives the proper validation error.
2. **Consider `validationContext: ['groups' => ['Default', 'Create']]` on the `CQRSCreate`** to match the module convention (and future-proof any Create-group constraints). Low impact today since your current constraints are in the Default group, but it's the documented pattern.

mattgoud (approved):
Thanks, validationContext ['Default','Create'] and Assert\NotBlank on cartRuleType are in. LGTM.

PR #325, opened 2026-07-06
Add root category endpoints to the Category domain

mattgoud:
| Questions         | Answers
| Branch?           | dev
| Description?      | Exposes the two missing root Category CQRS operations in the Admin API: `AddRootCategoryCommand` (`POST /categories/roots`) and `EditRootCategoryCommand` (`PATCH /categories/roots/{categoryId}`), both under the `category_write` scope.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Contributes to PrestaShop/PrestaShop#41973
| How to test?      | Run `CategoryEndpointTest`. `testAddRootCategory` / `testEditRootCategory` create and edit a root category via `/categories/roots`; `testEditUnknownRootCategoryReturnsNotFound` asserts the 404 mapping; `testInvalidAddRootCategory` / `testInvalidPatchRootCategory` assert the 422 validation path.
| Possible impacts? | New endpoints only; no change to existing operations.

Notes:
- The root add/edit operations live on the existing `Category` resource (same `GetCategoryForEditing` output and field set) with dedicated command mappings, because `AddRootCategoryCommand` takes `name`/`linkRewrite` constructor arguments and has no parent category (unlike `AddCategoryCommand`).
- The `UpdateCategoryPositionCommand` operation was **removed** from this PR and split out to PrestaShop/PrestaShop#42104: exposing the raw legacy `way`/`foundFirst`/`tr_` fields over REST lets a client corrupt sibling positions (reordering root categories with `foundFirst: true` produces duplicate positions). A clean contract needs a core-side self-normalizing position command. PR #331 (also position) is already closed.


mattgoud:
Thanks for the review. Triage of the points:

**Applied (e43bdf3):**
- `testAddRootCategory` now also asserts `enabled`.
- Added `testInvalidUpdateCategoryPosition` — sends a `positions` list that does not contain the moved category, which makes the handler throw `CategoryException` → asserts `422`, so the exception mapping is now covered.
- Clarified the `positions` docblock (the leading row-id segment is a legacy grid artifact and is ignored, any prefix works).
- Added a comment on `ADD_ROOT_COMMAND_MAPPING` explaining the `enabled` → `isActive` mapping.

**Verified, not a bug — `enabled` → `isActive`:**
`AddRootCategoryCommand(array $name, array $linkRewrite, $isActive)` and `EditRootCategoryCommand::setIsActive()` both use `isActive`, so the mapping is correct. The `[enabled] => [isEnabled]` in the regular `COMMAND_MAPPING` is actually a no-op on those commands (they expose `isActive`, not `isEnabled`), which is why the regular create relies on sending `isActive` directly. The green `testAddRootCategory` (which sends `enabled: true` and creates an active category) confirms the mapping end-to-end.

**Not applied, with reasons:**
- **Suggestion #2 (use `CategoryConstraintException` instead of `CategoryException` on the position endpoint):** the position handler throws the base `CategoryException` (`UpdateCategoryPositionHandler`: `throw new CategoryException('Category position cannot be updated')`), not `CategoryConstraintException`. Mapping `CategoryException → 422` is required here; switching to `CategoryConstraintException` would turn that failure into a 500. Kept as is.
- **Suggestion #5 (send `enabled` instead of `isActive` in the position test helper):** `AddCategoryCommand`'s constructor argument is `isActive`; `enabled` only maps to `isEnabled`, which that command does not expose. Sending `enabled` would leave `isActive` unset and break creation. `isActive` is the correct field (matching the pre-existing `createTemporaryCategories`).
- **Suggestion #6 (assert the resulting order after reorder):** intentionally omitted — the position is not returned by the API and asserting it reliably is entangled with the multishop `category_shop` position mechanism. The reordering behaviour itself is covered by the core Behat scenarios in `category_management.feature`; the endpoint here is covered by the happy-path (204) and the new 422 test.

tleon (dismissed):

nicosomb (dismissed):

SiraDIOP (commented):
Hello @mattgoud, 

Scenario: Create a root category via POST /categories/roots

Sent a POST request to /admin-api/categories/roots with names, linkRewrites, enabled: true, and shopIds: [1].

Received a 201 Created response with categoryId: 10, enabled: true, correct names and linkRewrites mapped to both configured locales (en-US, fr-FR), and parentCategoryId: 1 (the absolute root), as expected.

Checked the back office (Catalog > Categories): the new category "Ma root category test" appears correctly as a root-level category, alongside "Home", enabled and with the correct position.

Image

One note: the response body doesn't include shopIds, even though it was sent in the request. Not necessarily a bug since it may just not be part of the GetCategoryForEditing output, but flagging it in case it should be verified against the intended API contract.

Image

Scenario: POST /categories/roots without required names field

Sent a POST request without the names field.

Received a 400 Bad Request instead of the expected 422, with the following detail: MissingConstructorArgumentsException — the serializer fails before reaching validation, since AddRootCategoryCommand's constructor requires $name. The response also exposes a full stack trace including internal server file paths (/var/www/html/vendor/symfony/...) and class names.

⚠️ This isn't covered by a testInvalidAddRootCategory test yet. Recommend adding proper validation constraints so this returns a clean 422 instead of a raw serializer exception, and confirming stack traces aren't exposed this way outside of dev environments.

Image

Scenario: PATCH /categories/roots/{categoryId} without names (partial update)

Sent a PATCH request without names. Received a 200 OK — names remained unchanged while linkRewrites was updated as expected. Correct behavior for a partial update (CQRSPartialUpdate).

✅ Working as expected.

Image

Scenario: PATCH /categories/roots/{categoryId} with unknown categoryId

Sent a PATCH request to /categories/roots/999999. Received a 404 Not Found — CategoryNotFoundException correctly mapped, matching testEditUnknownRootCategoryReturnsNotFound.

Two things worth noting:

The response body's internal "status" field says 500, which is inconsistent with the actual HTTP status (404). Could be confusing for API clients reading that field.
Same stack trace exposure as above (internal file paths, handler class names).

✅ 404 mapping is correct, but the body inconsistency and stack trace exposure look like they could use a look — possibly a pre-existing pattern across the module rather than something introduced by this PR, worth confirming.

Image

Voici juste le commentaire pour ce dernier test (endpoint position) :

---

**Scenario: PUT /categories/{categoryId}/positions — reorder verification**

Sent a PUT request with `parentCategoryId: 1`, `way: 1`, `positions: ["tr_1_2", "tr_1_10"]`, `foundFirst: true`. Received 204 No Content. Back office order matched the request (Home = position 1, "Ma root category test" = position 2).

To verify the reorder actually applies (rather than just returning the request order by coincidence), sent the reverse order: `positions: ["tr_1_10", "tr_1_2"]`, `foundFirst: true`. Received 204 again, but the back office now shows **both categories at position 1** (duplicate), and the display order did not change to match the request (Home is still listed first).

🔴 **Bug**: the endpoint returns 204 (success) but the resulting sibling positions end up duplicated/inconsistent instead of a clean reorder. This looks related to the `foundFirst` parameter flagged in the AI pre-review — worth checking `UpdateCategoryPositionHandler` behavior when `foundFirst: true` is combined with a reversed `positions` list.

Also noticed: when the payload sends an invalid type (e.g. `foundFirst` as an int instead of bool), the response is a raw `NotNormalizableValueException` with a full stack trace, rather than a clean 422 — same pattern as seen on the POST endpoint.

Image

Image

Did i miss something?

mattgoud:
hey @SiraDIOP, thanks a lot for the thorough testing 🙏 quick triage:

**1. `shopIds` not in the response** — expected, `GetCategoryForEditing` doesn't return it, so it's not in the output contract. not a bug.

**2. POST without `names` → 400 / stack trace, and 404 body showing `status:500`** — not introduced by this PR: the already-merged `POST /categories` behaves identically (missing required field → serializer 400; missing default-language value → 500 "Invalid language data"). it's a module-wide validation-layer limitation, and the stack traces + `status:500` body are **dev-only** (in prod the bodies are clean and `status` matches the HTTP code). worth a separate issue for the missing→422 gap rather than fixing it here.
I did add `testInvalidAddRootCategory` + `testInvalidPatchRootCategory` covering the 422 path (forbidden char), which is the validation contract that actually works.

**3. PATCH partial update / 404 mapping** — 👍 confirmed working as intended.

**4. position reorder duplicate** — good catch, reproduced. it's specific to reordering **root** categories with `foundFirst: true`, which skips the core `cleanPositions()` (root positions live in `category_shop`); `foundFirst: false` normalizes correctly. that's the raw legacy param being exposed directly, so rather than shipping a corruptible endpoint I've **removed the position operation from this PR** and split it into PrestaShop/PrestaShop#42104 (needs a core-side self-normalizing command). #331 was already closed. this PR now only adds the root add/edit endpoints.

nothing missed on your side — really useful pass 🙌

mattgoud:
follow-up on point 2: opened #392 to track the missing/empty-required-field → 400/500-instead-of-422 behaviour module-wide (it's not specific to this PR).

PR #326, opened 2026-07-07
Add order remove-discount endpoint

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?       | Fills the `DeleteCartRuleFromOrderCommand` gap: `DELETE /orders/{orderId}/discounts/{orderCartRuleId}` (scope `order_write`) to remove a discount from an order.
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | `DELETE /orders/{id}/discounts/{orderCartRuleId}` (client granted `order_write`) removes the order discount. Covered by `OrderDiscountRemovalEndpointTest`.
| Possible impacts? | Removes a cart rule (discount) from the order and recalculates its totals.

### DIFF
--- a/src/ApiPlatform/Resources/Order/OrderDiscountRemoval.php
+++ b/src/ApiPlatform/Resources/Order/OrderDiscountRemoval.php
@@ -1,0 +1,54 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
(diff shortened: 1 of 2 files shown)


mattgoud (changes requested):
The endpoint itself is correct, but it needs a not-found test before it can be approved. `exceptionToStatus` maps `OrderNotFoundException => 404`, yet nothing exercises that path. Please add one, e.g.:

```php
public function testDeleteFromNonExistentOrderReturns404(): void
{
    $this->deleteItem('/orders/999999/discounts/999999', ['order_write'], Response::HTTP_NOT_FOUND);
}
```

While you're in the test, two small alignments with the module conventions: use the `deleteItem` helper instead of `requestApi`, and restore state via `DatabaseDump::restoreTables(['order_cart_rule'])` in setUp/tearDown rather than manual SQL (it also runs when a test fails mid-way).

For the record, you can disregard two points from the AI pre-review:
- **`OrderException => 422` "too broad"** is correct here — `DeleteCartRuleFromOrderHandler` throws the base `OrderException` directly (e.g. "Invalid order cart rule provided."), so narrowing to a specific `ConstraintException` would let those escape as 500.
- **The two `#[ApiProperty(identifier: true)]`** are appropriate for a two-path-variable URI (`/orders/{orderId}/discounts/{orderCartRuleId}`) — both variables must map to the command. The single-identifier precedents it cites only have one id in the path.

Once the 404 test is in, happy to approve.

mattgoud (approved):
Re-reviewed 👍 404 test added, switched to the `deleteItem` helper, and state is restored via `DatabaseDump::restoreTables(['order_cart_rule'])` in setUp/tearDown. LGTM.

Non-blocking nit for before merge: the `OrderDiscountRemoval` resource is missing `declare(strict_types=1)` — added it on #319, worth keeping the Order resources consistent.

PR #327, opened 2026-07-07
Add order duplicate-cart endpoint

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?       | Fills the `DuplicateOrderCartCommand` gap: `PUT /orders/{orderId}/cart-duplications` (scope `order_write`) to duplicate an order's cart. Returns the new cart id along with the source `orderId`.
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | `PUT /orders/{id}/cart-duplications` (client granted `order_write`) returns `{ orderId, cartId }` with the new cart id. Covered by `OrderCartDuplicationEndpointTest`.
| Possible impacts? | Creates a new cart duplicated from the order's cart.

### DIFF
--- a/src/ApiPlatform/Resources/Order/OrderCartDuplication.php
+++ b/src/ApiPlatform/Resources/Order/OrderCartDuplication.php
@@ -1,0 +1,53 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\Order;
+
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ApiResource;
+use PrestaShop\PrestaShop\Core\Domain\Order\Command\DuplicateOrderCartCommand;
+use PrestaShop\PrestaShop\Core\Domain\Order\Exception\DuplicateOrderCartException;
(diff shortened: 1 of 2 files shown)


mattgoud (changes requested):
`OrderNotFoundException => 404` is missing from `exceptionToStatus`. Only `DuplicateOrderCartException => 422` is mapped, so a bad `orderId` currently surfaces as 422 rather than a clean 404. Order endpoints should map the not-found case to 404.

Also missing a not-found / invalid negative test.

PrestaEdit:

- Added `OrderNotFoundException::class => Response::HTTP_NOT_FOUND` to `exceptionToStatus`, exactly as flagged. A bad `orderId` no longer falls through to 422; it returns 404.
- Added `testDuplicateNotFoundOrderCart` — PUT `/orders/99999/cart-duplications` asserts 404.

Ready for another look.

mattgoud:
The `OrderNotFoundException => 404` mapping is in place now, but it never gets reached and CI is red on the new `testDuplicateNotFoundOrderCart`. For an order with no cart, the core handler blows up first:

```
TypeError: PrestaShop\PrestaShop\Adapter\ContextStateManager::setCart():
Argument #1 ($cart) must be of type ?Cart, false given,
called in src/Adapter/Order/CommandHandler/DuplicateOrderCartHandler.php on line 47
```

`Cart::getCartByOrderId()` returns `false` and is passed straight to `setCart()`, so a bad orderId yields a 500, not a 404. This can't be fixed from the module: the core handler needs a guard (e.g. throw `OrderNotFoundException` / `DuplicateOrderCartException` when the cart is missing). I'd suggest a small core PR for the guard, then re-point the negative test at it.

PrestaEdit:
Good catch @mattgoud — confirmed, `Cart::getCartByOrderId()` returns `false` and the handler feeds it straight into `ContextStateManager::setCart(?Cart)`, so the `OrderNotFoundException → 404` mapping never gets a chance.

Opened a core PR to add the guard: PrestaShop/PrestaShop#42065 (throws `OrderNotFoundException` when the cart is missing).

In the meantime I marked `testDuplicateNotFoundOrderCart` as skipped here (with a reference to that PR) so CI goes green. I'll un-skip it as soon as the core guard lands.

mattgoud:
`OrderNotFoundException => 404` mapping is correct now 👍 — but I'm holding the changes-requested until core PrestaShop/PrestaShop#42065 lands: without that guard, a bad `orderId` hits the `TypeError` in `DuplicateOrderCartHandler` (feeds `false` into `setCart(?Cart)`) and returns 500 before the 404 mapping can fire, which is why the not-found test is skipped. happy-path is good. once #42065 is merged, unskip the test and I'll approve.

PrestaEdit:
Thanks for the review! Noting that PrestaShop/PrestaShop#42065 is open and mergeable (currently blocked on reviews/CI, last updated 2026-07-15). I'll unskip the not-found test and ping you here as soon as it lands.

PR #328, opened 2026-07-07
Add order change-currency endpoint

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?       | Fills the `ChangeOrderCurrencyCommand` gap: `PUT /orders/{orderId}/currencies` (scope `order_write`) to change an order's currency (`newCurrencyId`). Only allowed on unpaid orders — the handler rejects a change when the order is valid.
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | `PUT /orders/{id}/currencies` with `{ "newCurrencyId": x }` (client granted `order_write`) changes the currency of an unpaid order. Covered by `OrderCurrencyEndpointTest`.
| Possible impacts? | Changes the order currency and recalculates its totals.

### DIFF
--- a/src/ApiPlatform/Resources/Order/OrderCurrency.php
+++ b/src/ApiPlatform/Resources/Order/OrderCurrency.php
@@ -1,0 +1,55 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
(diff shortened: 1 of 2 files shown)


mattgoud (changes requested):
The mapping and exception handling look correct, but there's no negative test: the handler throws `OrderException` (422) when the order is already valid/paid — that's exactly the case worth asserting (the current setUp deliberately forces `valid=0` to avoid it). Please add a `testInvalid*` (422) covering it, per the module testing conventions.

PrestaEdit:
Thanks @mattgoud — addressed in the last push. Went with two negative tests to cover both paths:

- Added `#[Assert\Positive]` on `$newCurrencyId` — same shape as [#316](https://github.com/PrestaShop/ps_apiresources/pull/316) / [#317](https://github.com/PrestaShop/ps_apiresources/pull/317) / [#319](https://github.com/PrestaShop/ps_apiresources/pull/319).
- `testChangeOrderCurrencyWithInvalidPayload` — validator path: PUTs with `newCurrencyId: 0`, asserts 422 via `assertValidationErrors(...)`.
- `testChangeOrderCurrencyOnValidOrderReturns422` — handler path (the one you specifically flagged): temporarily flips the order to `valid=1`, expects the handler-thrown `OrderException` to surface as 422, restores `valid=0` in a `finally`. The response body isn't the Symfony validator shape here, so I asserted the status alone rather than `assertValidationErrors`.

Ready for another look.

mattgoud (approved):
Thanks, the 422 negative test is in and CI is green. LGTM.

PR #329, opened 2026-07-07
Add customer address-information endpoint to the Admin API

mattgoud:
| Questions         | Answers
| Branch?           | dev
| Description?      | Exposes the `GetCustomerForAddressCreation` query in the Admin API: `GET /customers/address-informations?email=...`, under the `customer_read` scope. It looks a customer up by email and returns the identity fields used to pre-fill an address form (`customerId`, `firstName`, `lastName`, `company`), mirroring the legacy address-creation autocomplete.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Related to PrestaShop/PrestaShop#41986
| How to test?      | Run `CustomerAddressInformationEndpointTest`. `testGetCustomerAddressInformation` creates a customer and fetches `/customers/address-informations?email=<email>`, asserting the full response; `testGetAddressInformationForUnknownEmailReturnsNotFound` asserts the 404 mapping for an unknown email.
| Possible impacts? | New read endpoint only; no change to existing operations.

Notes:
- The lookup key (email) is passed as a **query parameter**, not a path segment: an email in the path collides with API Platform's optional `.{_format}` suffix (the `.com` TLD would be parsed as a format) and needs awkward escaping of `@`/`.`. The `QueryProvider` merges query-string filters into the CQRS query construction, so `email` maps cleanly to the `GetCustomerForAddressCreation` constructor via `CQRSQueryMapping`.
- Modelled as a single-item `CQRSGet` (not a collection): the query returns one `AddressCreationCustomerInformation` DTO and throws when no customer matches. Since the URI template has no variable, API Platform resolves the operation with empty `uriVariables`; the `customerId` identifier is only used for the response contract.
- Only `CustomerByEmailNotFoundException → 404` is mapped. The generic `CustomerException` the handler can raise on a DB failure is intentionally left unmapped so it surfaces as a 500 rather than being masked as a 4xx.
- This covers one of the two endpoints listed in the issue. The `GetRequiredFieldsForCustomer` read endpoint is handled separately: it shares the `/customers/required-fields` resource introduced by #237 (the write side) and requires a small core change first (the CQRS serializer only wraps scalar query results under `_queryResult`, not list results).

### DIFF
--- a/src/ApiPlatform/Resources/Customer/CustomerAddressInformation.php
+++ b/src/ApiPlatform/Resources/Customer/CustomerAddressInformation.php
@@ -1,0 +1,82 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\Customer;
+
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ApiResource;
+use ApiPlatform\Metadata\Parameters;
+use ApiPlatform\Metadata\QueryParameter;
+use PrestaShop\PrestaShop\Core\Domain\Customer\Exception\CustomerByEmailNotFoundException;
+use PrestaShop\PrestaShop\Core\Domain\Customer\Query\GetCustomerForAddressCreation;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSGet;
+use Symfony\Component\HttpFoundation\Response;
+
+#[ApiResource(
+    operations: [
+        new CQRSGet(
+            uriTemplate: '/customers/address-informations',
+            CQRSQuery: GetCustomerForAddressCreation::class,
+            scopes: [
+                'customer_read',
+            ],
+            CQRSQueryMapping: self::QUERY_MAPPING,
+            parameters: new Parameters([
+                new QueryParameter(
+                    key: 'email',
+                    required: true,
+                    description: 'Email of the customer whose information is fetched to pre-fill an address (matches the first customer found for that email)'
+                ),
+            ]),
+            openapiContext: [
+                'parameters' => [
+                    [
+                        'name' => 'email',
+                        'in' => 'query',
+                        'required' => true,
+                        'schema' => [
+                            'type' => 'string',
+                        ],
+                        'description' => 'Email of the customer whose information is fetched to pre-fill an address',
+                    ],
+                ],
+            ],
+        ),
+    ],
+    normalizationContext: ['skip_null_values' => false],
+    exceptionToStatus: [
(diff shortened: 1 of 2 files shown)


mattgoud:
Thanks for the pass. The three "needs verification against Core" points are all confirmed working — both by reading the framework and by the 6 green integration jobs (8.1/8.4/8.5 × develop/9.0.3/9.1.x):

- **`CQRSGet` without a URI variable** — `UriTemplateResourceMetadataCollectionFactory::configureUriVariables()` resolves a template with no path variable to empty `uriVariables`, so `identifier: true` on `customerId` is simply unused for routing (it only shapes the response). No error, and `testGetCustomerAddressInformation` exercises the full route.
- **`CQRSQueryMapping` used as input** — it is not output-only. `QueryProvider` applies it as the `NORMALIZATION_MAPPING` when denormalizing the request (`uriVariables + filters + context`) into the query, *and* when normalizing the result. `[email] => [customerEmail]` feeds the query constructor (input) and is a no-op on the result (which has no `email`). The passing test proves the email reaches the query.
- **`CustomerId` VO unwrapping** — `ValueObjectNormalizer` normalizes a VO inline to its scalar (`getCustomerId()` → `customerId: <int>`), so no explicit mapping is needed. The test asserts the exact int and passes.

On the two suggestions:

- **URI singular `address-information`** — not possible: `ApiResourceUriTemplateRector` (CI-enforced) requires a plural last segment and fails on the singular. Keeping `address-informations`, which also matches the domain vocabulary (`AddressCreationCustomerInformation`, `personalInformation`, …).
- **`testInvalid*` for a missing `email`** — the `testInvalid*` convention targets create/update body validation. No lookup endpoint in the module (`FoundCustomer`, `FoundProduct`, `FoundCartRule`) tests a missing `required` query parameter; the happy path plus the 404 (`CustomerByEmailNotFoundException`) cover this endpoint's contract.

No code changes from the review.

nicosomb (approved):

src/ApiPlatform/Resources/Customer/CustomerAddressInformation.php
```diff
@@ -0,0 +1,82 @@
...
+use ApiPlatform\Metadata\QueryParameter;
+use PrestaShop\PrestaShop\Core\Domain\Customer\Exception\CustomerByEmailNotFoundException;
+use PrestaShop\PrestaShop\Core\Domain\Customer\Query\GetCustomerForAddressCreation;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSGet;
+use Symfony\Component\HttpFoundation\Response;
+
+#[ApiResource(
+    operations: [
+        new CQRSGet(
+            uriTemplate: '/customers/address-informations',
```
jolelievre, line 35:
```suggestion
            uriTemplate: '/customers/search',
```
The query is more about searching a customer by email, that's the only reason why it's relevant compared to existing endpoints that return much more date, even if in the BO the use case is purely limited to searching a customer for the address creation, I think we can afford being a bit more generic here

src/ApiPlatform/Resources/Customer/CustomerAddressInformation.php
```diff
@@ -0,0 +1,82 @@
...
+                ],
+            ],
+        ),
+    ],
+    normalizationContext: ['skip_null_values' => false],
+    exceptionToStatus: [
+        CustomerByEmailNotFoundException::class => Response::HTTP_NOT_FOUND,
+    ],
+)]
+class CustomerAddressInformation
```
jolelievre, line 68:
```suggestion
class FoundCustomer
```

tests/Integration/ApiPlatform/CustomerAddressInformationEndpointTest.php
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace PsApiResourcesTest\Integration\ApiPlatform;
+
+use Symfony\Component\HttpFoundation\Response;
+use Tests\Resources\DatabaseDump;
+
+class CustomerAddressInformationEndpointTest extends ApiTestCase
```
jolelievre, line 28:
```suggestion
class FoundCustomerEndpointTest extends ApiTestCase
```

cnavarro-prestashop:
After discussion with the team : another endpoint covers all data (`/customers/search`) returned by `/customers/address-informations` so it's useless to expose : we'll ignore it in Admin API tracking table.

PR #336, opened 2026-07-08
Add order preview read endpoint

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?       | Exposes `GetOrderPreview` as `GET /orders/{orderId}/preview` (scope `order_read`): returns the invoice and shipping details, product details, tax/virtual flags and the formatted addresses of an order. Lightweight alternative to the full `/details` endpoint.
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | `GET /orders/{id}/preview` (client granted `order_read`) returns `{ orderId, invoiceDetails, shippingDetails, productDetails, taxIncluded, virtual, invoiceAddressFormatted, shippingAddressFormatted }`. A non-existent id returns 404. Covered by `OrderPreviewEndpointTest`.

### DIFF
--- a/src/ApiPlatform/Resources/Order/OrderPreview.php
+++ b/src/ApiPlatform/Resources/Order/OrderPreview.php
@@ -30,7 +30,7 @@ use Symfony\Component\HttpFoundation\Response;
 #[ApiResource(
     operations: [
         new CQRSGet(
-            uriTemplate: '/orders/{orderId}/previews',
+            uriTemplate: '/orders/{orderId}/preview',
             requirements: ['orderId' => '\d+'],
             CQRSQuery: GetOrderPreview::class,
             scopes: [
--- a/src/ApiPlatform/Resources/Order/OrderPreview.php
+++ b/src/ApiPlatform/Resources/Order/OrderPreview.php
@@ -30,7 +30,7 @@ use Symfony\Component\HttpFoundation\Response;
 #[ApiResource(
     operations: [
         new CQRSGet(
-            uriTemplate: '/orders/{orderId}/preview',
+            uriTemplate: '/orders/{orderId}/previews',
             requirements: ['orderId' => '\d+'],
             CQRSQuery: GetOrderPreview::class,
             scopes: [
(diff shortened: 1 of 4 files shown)


mattgoud (changes requested):
The resource is clean (the earlier pre-review findings about deprecated `OrderPreview*` sub-objects and the dead `QUERY_MAPPING` are resolved — the DTO exposes only the modern formatted strings, no Value Objects). One point to settle before approving:

**URI singular vs plural.** The `uriTemplate` is `/orders/{orderId}/previews` (plural) but the PR description says `/orders/{orderId}/preview` (singular). An order has exactly one preview — a singleton sub-resource — so per the `/categories/{categoryId}/cover` precedent it should be singular: `/orders/{orderId}/preview`. Please align the `uriTemplate` (and the description) on the singular form.

Minor, while you're there: the test only asserts key presence/types — asserting a couple of concrete values (e.g. `invoiceAddressFormatted`/`shippingAddressFormatted` non-empty, `productDetails` non-empty) would make it catch serialization regressions. Not a blocker.

Once the URI is settled, happy to approve.

PrestaEdit:
Thanks for the review @mattgoud. Aligned on the singular `/orders/{orderId}/preview` per the `/cover` precedent — a prior revision had reverted it to plural because the `ApiResourceUriTemplateRector` rule doesn't know that `preview` is a singleton sub-resource. Fixed both the URI (resource + test) and added `preview` to the Rector rule's `SKIPPED_KEYWORDS` (with a fixture) so it stays singular. Test already asserts non-empty on the formatted addresses and `productDetails` — matches your minor suggestion.

mattgoud (approved):
Re-reviewed 👍 URI is now the singular `/orders/{orderId}/preview` (+ the Rector `preview` keyword and skip fixture, nice touch), the DTO exposes only the formatted strings, and the 404 test is in place. LGTM.

Same non-blocking nit as #326: `OrderPreview` is missing `declare(strict_types=1)`.

PR #337, opened 2026-07-08
Add product attribute groups read endpoint

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?       | Exposes `GetProductAttributeGroups` as `GET /products/{productId}/attribute-groups` (scope `product_read`): returns the attribute groups (with their nested attributes) associated with a product's combinations.
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | `GET /products/{id}/attribute-groups` (client granted `product_read`) returns the product's attribute groups (with nested attributes). Covered by `ProductAttributeGroupListEndpointTest`.

### DIFF
--- a/src/ApiPlatform/Resources/Product/ProductAttributeGroupList.php
+++ b/src/ApiPlatform/Resources/Product/ProductAttributeGroupList.php
@@ -1,0 +1,63 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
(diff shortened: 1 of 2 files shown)


mattgoud (commented):
The red integration legs all fail on the same assertion, `testGetAttributeGroupsForNonExistentProduct` (expects 404), across every core version — so it's a real contract issue, not a version gap or flake.

Root cause: the resource maps `ProductNotFoundException → 404`, but the wrapped query `GetProductAttributeGroups` never throws it. `GetProductAttributeGroupsHandler::handle()` returns `[]` when the product has no attribute-group ids, and `ProductRepository::getProductAttributesGroupIds()` is a plain SELECT with no existence check — so an unknown product returns **200 `[]`**, the `exceptionToStatus` mapping is dead code, and the test can never pass.

Also note a *real* product with no combinations returns `[]` too, so "missing product" and "product without attributes" aren't distinguishable at the query level.

Two ways forward:
- **Keep the 404** → needs a small core change: call the existing `assertProductExists()` in `GetProductAttributeGroupsHandler` (or in the repository method). This makes the mapping + test valid. Blocked on that core PR.
- **Module-only** → drop the `exceptionToStatus` mapping and the 404 test, and document that an unknown/combination-less product yields `200 []`.

Happy-path mapping is otherwise correct.

jolelievre:

Note: the unreachable 404 mapping was dropped in #410 (the core query returns an empty list for unknown products, so the endpoint documents the 200 [] contract instead), and the nested attribute names are now keyed by locale.

PR #338, opened 2026-07-08
Add product catalog price rules read endpoint

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?       | Exposes `GetCatalogPriceRuleListForProduct` as `GET /products/{productId}/catalog-price-rules` (scope `catalog_price_rule_read`): returns the catalog price rules that apply to a product with the total count.
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | `GET /products/{id}/catalog-price-rules` (client granted `catalog_price_rule_read`) returns `{ productId, catalogPriceRules, totalCount }`. Covered by `ProductCatalogPriceRulesEndpointTest`.

### DIFF
--- a/src/ApiPlatform/Resources/Product/ProductCatalogPriceRules.php
+++ b/src/ApiPlatform/Resources/Product/ProductCatalogPriceRules.php
@@ -48,7 +48,7 @@ class ProductCatalogPriceRules
     public int $totalCount;
 
     public const QUERY_MAPPING = [
-        '[_context][langId]' => '[languageId]',
+        '[_context][langId]' => '[langId]',
         '[productId]' => '[productId]',
     ];
 }
--- a/src/ApiPlatform/Resources/Product/ProductCatalogPriceRules.php
+++ b/src/ApiPlatform/Resources/Product/ProductCatalogPriceRules.php
@@ -1,0 +1,54 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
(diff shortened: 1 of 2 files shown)


PrestaEdit:
Thanks for the review. Pushed a follow-up commit addressing most points:

- **`declare(strict_types=1)`** added.
- **`ApiProperty(openapiContext: …)`** added on `$catalogPriceRules`, describing the `CatalogPriceRuleForListing` shape (id, name, quantity, reduction fields, dates, denormalized shop/currency/country/group names).
- **Test cleanup**: removed the redundant `self::createApiClient(...)` in `setUpBeforeClass` and tightened assertions (types + `totalCount === count(catalogPriceRules)`).
- **QUERY_MAPPING output** — verified: the Core query handler returns a `CatalogPriceRuleList` whose getters are `getCatalogPriceRules()` / `getTotalCount()`, matching the DTO property names, so no output mapping is needed.

Skipped: **`exceptionToStatus` + `testInvalid…` 404**. The `GetCatalogPriceRuleListForProductHandler` does not throw `ProductNotFoundException` — it just calls `CatalogPriceRuleRepository::getByProductId/countByProductId`, which return an empty list / 0 for an unknown product id. So a request to `/products/99999/catalog-price-rules` today returns 200 with `{catalogPriceRules: [], totalCount: 0}`, not 500. Adding an `exceptionToStatus` entry for an exception the handler never throws, and a `testInvalid…` expecting 404, would be misleading. Happy to change the handler behavior (validate product existence up front) in a separate Core PR if maintainers prefer strict 404 semantics — but that's out of scope for this API-Platform-only PR.

Scope choice (`catalog_price_rule_read` vs `product_read`): opted for `catalog_price_rule_read` because the returned entity is a list of catalog price rules — the product is only the filter. Open to switching if maintainers prefer sub-resources to gate on the parent scope.

PR #339, opened 2026-07-08
Add carrier ranges endpoint

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?       | Fills the `SetCarrierRangesCommand` gap: `PUT /carriers/{carrierId}/ranges` (scope `carrier_write`) to set a carrier's shipping ranges. Each range is `{ id_zone, range_from, range_to, range_price }`. Standalone resource file — doesn't collide with the other Carrier PRs.
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | `PUT /carriers/{id}/ranges` with `{ "ranges": [ { "id_zone": z, "range_from": 0, "range_to": 100, "range_price": "10.00" } ] }` (client granted `carrier_write`) sets the carrier ranges. Covered by `CarrierRangesEndpointTest`.
| Possible impacts? | May create a new carrier version (as the domain handler does when needed) and updates the carrier's price ranges.

### DIFF
--- a/src/ApiPlatform/Resources/Carrier/CarrierRanges.php
+++ b/src/ApiPlatform/Resources/Carrier/CarrierRanges.php
@@ -1,0 +1,65 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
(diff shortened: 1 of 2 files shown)


PrestaEdit:
Closing: like `SetCarrierTaxRuleGroupCommand` (dropped from #302 for the same reason), `CarrierRangeRepository::set` also asserts `shopConstraint->forAllShops()` (`src/Adapter/Carrier/Repository/CarrierRangeRepository.php:52,80,144,211`), which the API request context does not provide. This endpoint would need a core change to allow single-shop constraints (or a way to send an all-shops constraint through the Admin API) before it can be exposed.

PR #340, opened 2026-07-08
Add employee email read endpoint

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?       | Exposes `GetEmployeeEmailById` as `GET /employees/{employeeId}/emails` (scope `employee_read`): returns `{ employeeId, email }`. The query result is a mono-arg `Email` VO that collapses to its scalar string, mapped from `_queryResult` to the resource's `email` property (same pattern as `ShowcaseCard`).
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | `GET /employees/{id}/emails` (client granted `employee_read`) returns `{ employeeId, email }`. A missing employee id returns 404. Covered by `EmployeeEmailEndpointTest`.

### DIFF
--- a/src/ApiPlatform/Resources/Employee/EmployeeEmail.php
+++ b/src/ApiPlatform/Resources/Employee/EmployeeEmail.php
@@ -1,0 +1,61 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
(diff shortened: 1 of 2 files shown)


tleon (approved):

jolelievre (changes requested):
I don't think this one is useful, simply returning the email based on an ID is a use case we may use in the BO, but I don't hink it's valuable for an API, the regular GET endpoint already returns all the employee details, including email

I know it was marked as an endpoint to handle, but that's because the issue was automatically generated based on all existing queries/commands We need to filter the ones that are really valuable and avoid useless doublons

Unless you see a use case where it's really useful for the API to return only the email, but I don't see it

PrestaEdit:
Agreed — the regular `GET /employees/{id}` already returns the email, so this endpoint is a redundant 1:1 mapping of the CQRS query without added API value. Closing. I'll pick a more meaningful endpoint from #39630 for the next PR.

PR #346, opened 2026-07-09
Add currency exchange rate read endpoint

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?       | Exposes `GetCurrencyExchangeRate` as `GET /currencies/exchange-rates?isoCode=...` (scope `currency_read`): returns the CLDR exchange rate for a currency ISO code. The handler returns an `ExchangeRate` value object wrapping a `DecimalNumber`; the framework's `_queryResult` wrapper is mapped to the `exchangeRate` property (round-trips as a string, same normalization as `CustomerGroup`'s decimal fields). Uses the CQRSGet + QueryParameter pattern.
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | `GET /currencies/exchange-rates?isoCode=USD` (client granted `currency_read`) returns `{ exchangeRate }`. An unknown code returns 404. Covered by `ExchangeRateEndpointTest`.

### DIFF
--- a/src/ApiPlatform/Resources/Currency/ExchangeRate.php
+++ b/src/ApiPlatform/Resources/Currency/ExchangeRate.php
@@ -1,0 +1,66 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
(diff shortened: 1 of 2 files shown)


PrestaEdit:
Closing: the `GetCurrencyExchangeRate` query returns an `ExchangeRate` VO wrapping a `DecimalNumber` (not a plain scalar). The framework's `_queryResult` wrapper only kicks in for scalar results (see `QueryResultSerializerTrait::denormalizeQueryResult` — `is_scalar` check), so the composite VO isn't wrapped and the mapping doesn't fire — the response omits the `exchangeRate` key. Would need either a core change to the trait (`is_scalar || $result instanceof ValueObject`) or a custom normalizer on this endpoint. Deprioritized for now.

PR #350, opened 2026-07-10
Add Carrier endpoints

### LINKED ISSUE #41972: Admin API - Add missing endpoints for the "Carrier" domain
Endpoints:
- AddCarrierCommand
- EditCarrierCommand 
- SetCarrierRangesCommand 
- SetCarrierTaxRuleGroupCommand
- GetAvailableCarriers 
- GetCarriersForProduct

nicosomb:
| Questions         | Answers
| Description?      | Add Carrier endpoints
| Type?             |  new feature
| BC breaks?        | no
| Deprecations?     |  no
| Fixed ticket?     | 
| How to test?      | CI and tests green


tleon (dismissed):

cnavarro-prestashop (dismissed):

src/ApiPlatform/Resources/Carrier/CarrierTaxRuleGroup.php
```diff
@@ -0,0 +1,63 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Carrier\Exception\CarrierNotFoundException;
+use PrestaShop\PrestaShop\Core\Domain\TaxRulesGroup\Exception\TaxRulesGroupNotFoundException;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSPartialUpdate;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Validator\Constraints as Assert;
+
+#[ApiResource(
+    operations: [
+        new CQRSPartialUpdate(
+            uriTemplate: '/carriers/{carrierId}/tax-rule-group',
```
jolelievre, line 37: This one should be integrated in the same ApiResource so we can return the full carrier as a full REST response

jolelievre (changes requested):

src/ApiPlatform/Resources/Carrier/CarrierTaxRuleGroup.php
jolelievre:
```suggestion
            uriTemplate: '/carriers/{carrierId}/set-tax-rule-group',
```
Rename more into an action, don't forget to add the exception in rector rule

src/ApiPlatform/Resources/Carrier/CarrierRanges.php
```diff
@@ -0,0 +1,99 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Carrier\Exception\CarrierConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Carrier\Exception\CarrierNotFoundException;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSPartialUpdate;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Validator\Constraints as Assert;
+
+#[ApiResource(
+    operations: [
+        new CQRSPartialUpdate(
+            uriTemplate: '/carriers/{carrierId}/ranges',
```
jolelievre, line 46: The get endpoints based on `GetCarrierRanges` should be integrated in this same API resource And also used as a CQRSQuery for the PATCH endpoint so it returns the updated state after modification

tests/Integration/ApiPlatform/AvailableCarriersEndpointTest.php
```diff
@@ -0,0 +1,233 @@
...
+            'address_format',
+            'customer',
+            'customer_group',
+            'address',
+        ]);
+    }
+
+    protected function setUp(): void
+    {
+        parent::setUp();
```
jolelievre: Prefer skipping based on the min version rather than the code, also if some modifications were required in 9.2.x, then a minVersion should be added on the related endpoints

tests/Integration/ApiPlatform/AvailableCarriersEndpointTest.php
```diff
@@ -0,0 +1,233 @@
...
+        $carrierId = $carrier['carrierId'];
+
+        $product = $this->createItem('/products', [
+            'type' => ProductType::TYPE_STANDARD,
+            'names' => ['en-US' => 'Available carrier product', 'fr-FR' => 'Produit disponible'],
+        ], ['product_write']);
+        $productId = $product['productId'];
+
+        // Carrier reference id equals carrier id right after creation (they only diverge on later edits).
+        $commandBus->handle(new SetCarriersCommand($productId, [$carrierId], ShopConstraint::allShops()));
```
jolelievre: This will be doable once the PR https://github.com/PrestaShop/ps_apiresources/pull/410 is merged, so we should wait until the endpoint is available for integration tests to be fully based on API

src/ApiPlatform/Resources/Carrier/AvailableCarriers.php
```diff
@@ -0,0 +1,108 @@
...
+use ApiPlatform\Metadata\QueryParameter;
+use PrestaShop\PrestaShop\Core\Domain\Address\Exception\AddressNotFoundException;
+use PrestaShop\PrestaShop\Core\Domain\Carrier\Query\GetAvailableCarriers;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSGet;
+use Symfony\Component\HttpFoundation\Response;
+
+#[ApiResource(
+    operations: [
+        new CQRSGet(
+            uriTemplate: '/carriers/available',
```
jolelievre:
```suggestion
            uriTemplate: '/carriers/search-available-carriers',
```
The description of this endpoint should be declared clearly, it seems like this query is used to search the available carrier for a restricted context:
- list of products (so carrier msut handle them)
- a defined address (so the carrier must be compatible with the country/state/zone)

It's an endpoint that is likely used when creating a cart so the context of usage is very specific I suggested this new naming but maybe it can be improved to reflect the specific usage of this search endpoint Maybe "compatible" would be more accurate

jolelievre (dismissed):

tests/Integration/ApiPlatform/CarrierListEndpointTest.php
```diff
@@ -0,0 +1,219 @@
...
+        $carriers = $this->listItems('/carriers', ['carrier_read']);
+        $this->assertGreaterThanOrEqual(1, $carriers['totalItems']);
+        // The list is ordered by position, which is also the order the carriers are offered in
+        $this->assertEquals('position', $carriers['orderBy']);
+
+        $filtered = $this->listItems('/carriers', ['carrier_read'], ['name' => 'Carrier for list']);
+        $this->assertEquals(1, $filtered['totalItems']);
+
+        $listedCarrier = $filtered['items'][0];
+        $this->assertEquals($carrierId, $listedCarrier['carrierId']);
```
jolelievre: No this should be replaced by a single assertEquals that test the chole array structure

jolelievre (dismissed):

jolelievre (dismissed):

mattgoud (dismissed):

mattgoud:
QA OK on `4decf50`

Tested as a pair with PrestaShop/PrestaShop#42022, on a fresh 9.2.0 Docker install of that branch with this module installed at `4decf50` through the core lock. Full report on the core side: PrestaShop/PrestaShop#42022 (comments of the 24th and of today).

Every carrier endpoint of this PR was exercised live:

| Endpoint | Result |
|---|---|
| `GET /carriers` | 200, list ordered by position |
| `POST /carriers`, defaults omitted | `additionalHandlingFee` false, `free` false, `shippingMethod` 2, `rangeBehavior` 0 |
| `POST /carriers`, values provided | never overwritten, including `free: true` |
| `POST /carriers` with `position` | followers shifted, no duplicate position |
| `POST /carriers` multipart with a logo | `img/s/{id}.jpg` and the generated thumbnail both served |
| `GET /carriers/{id}` | 200, unknown id 404 |
| `POST /carriers/{id}` | 200, only the submitted field changes |
| `PATCH /carriers/{id}/set-tax-rule-group` | 200 |
| `GET /carriers/{id}/ranges` | 200 |
| `PATCH /carriers/{id}/ranges` | 200, written to the table matching the method |
| `GET /carriers/search-compatible-carriers` | 200 on a valid payload, 400 on a missing, zero or negative quantity |
| `GET` and `PUT /products/{id}/carriers` | 200 |
| `PUT /products/99999/carriers` | 404, the product existence fix works |

Re-verified on `4decf50` after the removal of the shop configuration shipping method: creation without `shippingMethod` still returns `2`, and `shippingMethod: 0` is now rejected with `422 The value you selected is not a valid choice.`, since `Assert\Choice(choices: ShippingMethod::AVAILABLE_VALUES)` follows the core value object.

On the red CI here: the six failing carrier tests all depend on the core changes of PrestaShop/PrestaShop#42022, which this workflow cannot see because it checks out `develop` and `9.2.x` without it. The authoritative run is the "API Module tests" job of the core pull request, which installs this module into the core branch, and it is green there.

One point left, pre-existing and out of scope for this PR: `GET /products/{id}/carriers` answers 200 for an unknown product while the `PUT` on the same path answers 404, so a consumer probing with `GET` would conclude the product exists.

mattgoud:
Reopening one point on `4decf50`, found while replaying the product team's August feedback against the current state.

**An unknown zone identifier answers 500 instead of 404**, on creation and on update:

```
POST /admin-api/carriers        {"zones": [9999], ...}  -> 500
POST /admin-api/carriers/1      {"zones": [9999]}       -> 500

class : PrestaShop\PrestaShop\Core\Domain\Zone\Exception\ZoneNotFoundException
detail: zone #9999 does not exist
```

The domain check is there and does its job, the exception is raised. What is missing is the HTTP mapping: `exceptionToStatus` on the `Carrier` resource declares five exceptions and this one is not among them.

```php
exceptionToStatus: [
    CarrierNotFoundException::class => Response::HTTP_NOT_FOUND,
    CarrierConstraintException::class => Response::HTTP_UNPROCESSABLE_ENTITY,
    CannotAddCarrierException::class => Response::HTTP_UNPROCESSABLE_ENTITY,
    CannotUpdateCarrierException::class => Response::HTTP_UNPROCESSABLE_ENTITY,
    TaxRulesGroupNotFoundException::class => Response::HTTP_NOT_FOUND,
],
```

Adding `ZoneNotFoundException::class => Response::HTTP_NOT_FOUND` should be enough. `TaxRulesGroupNotFoundException` is already mapped that way, so the same treatment for zones keeps the resource consistent.

This was reported by the product team on the 14th of August, and answered at the time with "a check is missing, I am adding it". The check did land, only the status mapping did not follow.

My earlier QA comment stands for everything else: I had only exercised valid zone identifiers, so this case was outside my coverage. I am moving the label back to **Waiting for author** until this is addressed.

For the record, everything else from that feedback list now checks out on this head: the position is shifted instead of duplicated, `additionalHandlingFee`, `maxWidth`, `maxHeight`, `maxDepth` and `maxWeight` are optional and documented, the logo is handled, `associatedGroupIds: []` is rejected with a 422, `taxRuleGroupId` left the input schema, the ranges endpoint answers 200 with the right payload shape, `set-tax-rule-group` answers 200, and `GET /products/{id}/carriers` carries no pagination.

jolelievre (approved):

mattgoud (approved):

ps-jarvis:
PR merged, well done!

PR #352, opened 2026-07-11
Add GetProductIsEnabled Admin API endpoint

PrestaEdit:
| Description?      | Adds `GET /products/{productId}/status` using `CQRSGet` with `GetProductIsEnabled`. The bool return is exposed via the `[_queryResult]` scalar-wrap mapping as `{enabled: bool}`. |
| How to test?      | Run `composer phpunit-integration`; new `ProductIsEnabledEndpointTest` covers scope protection, GET happy path, and unknown-id → 404. |

Endpoint added:
- `GET /products/{productId}/status` — bool scalar-wrapped, `ProductNotFoundException → 404`.

Same scalar-wrap pattern as `GetShowcaseCardIsClosed` on `ShowcaseCard` (`[_queryResult]` → property).

### DIFF
--- a/src/ApiPlatform/Resources/Product/ProductIsEnabled.php
+++ b/src/ApiPlatform/Resources/Product/ProductIsEnabled.php
@@ -1,0 +1,54 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\Product;
+
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ApiResource;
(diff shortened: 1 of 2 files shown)


mattgoud:
Quick synthesis on top of the AI pre-review 👇

**False positive to discard:** the "missing shop constraint (likely blocker)" point. `GetProductIsEnabled::__construct(int $productId)` takes only a product id, there is no `ShopConstraint` argument to map. Adding `'[_context][shopConstraint]' => '[shopConstraint]'` would actually break denormalization. Nothing to change here, and CI (incl. happy path) is green.

**Worth doing (tests):**
- Assert the actual value of `enabled`, not just its type (the default catalog product 1 is enabled, so `assertTrue($result['enabled'])`).
- Move the 404 case to `getItem('/products/999999/status', ['product_read'], Response::HTTP_NOT_FOUND)` instead of raw `requestApi`.

**Minor:** the URI is `/products/{productId}/status` but the description says `/is-enabled`. Worth aligning the two (`/is-enabled` mirrors the query name and the `enabled` field).

PrestaEdit:
@mattgoud Applied the two test tweaks in the latest push:

- Happy path now asserts `assertTrue($result['enabled'])`.
- 404 case switched to `getItem('/products/999999/status', ['product_read'], Response::HTTP_NOT_FOUND)`.

On the URI/description mismatch: aligned the description to the code (`/status`) rather than the reverse. Earlier commits tried `/is-enabled` and `/enable-status` — both got rewritten by `ApiResourceUriTemplateRector` because `is-enabled` isn't in its `SKIPPED_KEYWORDS` list (whereas `status` is). Happy to revisit if we want to extend the Rector allowlist, but that feels out of scope for this PR.

mattgoud (approved):
Nits addressed 👍 the test now asserts `enabled === true` and uses `getItem` for the 404 case. The single red leg earlier was an infra flake at the Docker/apt setup step (not a test failure); re-run is green across the full matrix now. LGTM.

tleon (approved):

jolelievre:
Closing without integrating: GetProductIsEnabled is a small duplicate — GET /products/{productId} already returns the enabled status (and much more), so a dedicated endpoint is not relevant for the Admin API even though it serves a specific back office use case. It is now listed in GenerateApiTrackingTableCommand::EXCLUDED_CQRS_CLASSES (see #410, the PR that centralizes all the pending Product-domain endpoints) so it is excluded from the tracking table and won't be picked up again. @PrestaEdit thanks for the contribution — the review feedback you addressed here informed the conventions applied in #410.

PR #353, opened 2026-07-11
Add GetProductSupplierOptions Admin API endpoint

PrestaEdit:
| Description?      | Adds `GET /products/{productId}/supplier-options` using `CQRSGet` with `GetProductSupplierOptions`. Response shape: `defaultSupplierId`, `supplierIds[]`, and the `productSuppliers[]` list detailing per-supplier reference/price/currency/combination. |
| How to test?      | Run `composer phpunit-integration`; new `ProductSupplierOptionsEndpointTest` covers scope protection, GET happy path, and unknown-id → 404. |

Endpoint added:
- `GET /products/{productId}/supplier-options` — mono-arg productId, `ProductNotFoundException → 404`.

### DIFF
--- a/src/ApiPlatform/Resources/Product/ProductSupplierOptions.php
+++ b/src/ApiPlatform/Resources/Product/ProductSupplierOptions.php
@@ -1,0 +1,72 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+declare(strict_types=1);
+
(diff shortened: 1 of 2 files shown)


mattgoud:

**False positives to discard:**
- "Missing CQRSQueryMapping – critical": not needed. The query result `ProductSupplierOptions` and its nested `ProductSupplierForEditing` are `Core\Domain` objects, so the CQRS normalizer walks their getters, which already match the DTO property names; `productId` is echoed from the URI. No mapping required, and the happy-path leg is green.
- "Shop constraint not forwarded": correct as-is, the query is mono-arg (`productId` only), no `ShopConstraint`.
- `priceTaxExcluded` typed as string is fine (`DecimalNumber` serializes to a string).

**Worth doing (test):** the happy path is too shallow. It picks `MIN(id_product)`, which likely has no supplier, so `productSuppliers[]` is empty and its item shape is never asserted. Please target a product with known supplier data and assert the nested item fields (`reference`, `supplierId`, `currencyId`, `combinationId`, ...).

PrestaEdit:
Thanks @mattgoud! You were right on both points.

The shallow happy-path came from `MIN(id_product)`, which in the demo data is a combinations-type product — the handler short-circuits `productSuppliers` to `[]` in that case, so the nested shape was never exercised.

Fixed in the latest push: the test now picks the first standard product, seeds a supplier association via `SetSuppliersCommand` + `UpdateProductSuppliersCommand`, and asserts every field of `productSuppliers[0]` (`productSupplierId`, `productId`, `supplierId`, `supplierName`, `reference`, `priceTaxExcluded`, `currencyId`, `combinationId`).

mattgoud (approved):
Nit addressed 👍 the test now seeds a supplier association (`SetSuppliersCommand` + `UpdateProductSuppliersCommand`) and asserts the nested `productSuppliers[]` item fields against the seeded reference/price/currency, so the per-supplier shape is actually exercised. CI is green across the full matrix. LGTM.

tleon (approved):

jolelievre:

Note: the GET moved from /products/{productId}/supplier-options to /products/{productId}/suppliers in #410, where it shares its URI and resource with the supplier write operations.

PR #354, opened 2026-07-12
Add GetShopProductImages Admin API endpoint

PrestaEdit:
| Description?      | Adds `GET /products/{productId}/shop-images` using `CQRSGetCollection` with `GetShopProductImages`. Returns per-shop image associations: each element = `{shopId, productImages: [{imageId, cover}, ...]}`. Complements the existing `GET /products/{productId}/images` (`ProductImageList`, single-shop) with the multi-shop view. Unknown product id returns `200 []` (no exception), matching the underlying all-shops query. |
| How to test?      | Run `composer phpunit-integration`; `ShopProductImagesEndpointTest` covers scope protection and the list happy path, asserting the exact per-shop `{imageId, cover}` grouping against the fixture data (`image_shop` table). |

### DIFF
--- a/src/ApiPlatform/Resources/Product/ShopProductImages.php
+++ b/src/ApiPlatform/Resources/Product/ShopProductImages.php
@@ -43,7 +43,6 @@ use Symfony\Component\HttpFoundation\Response;
 )]
 class ShopProductImages
 {
-    #[ApiProperty(identifier: true)]
     public int $shopId;
 
     #[ApiProperty(openapiContext: [
--- a/src/ApiPlatform/Resources/Product/ShopProductImages.php
+++ b/src/ApiPlatform/Resources/Product/ShopProductImages.php
@@ -24,10 +24,8 @@ namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\Product;
 
 use ApiPlatform\Metadata\ApiProperty;
 use ApiPlatform\Metadata\ApiResource;
-use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductNotFoundException;
 use PrestaShop\PrestaShop\Core\Domain\Product\Image\Query\GetShopProductImages as GetShopProductImagesQuery;
 use PrestaShopBundle\ApiPlatform\Metadata\CQRSGetCollection;
-use Symfony\Component\HttpFoundation\Response;
 
 #[ApiResource(
     operations: [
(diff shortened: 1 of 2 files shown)


mattgoud:

**Overstated / false positives:**
- "Missing `exceptionToStatus` = hard blocker / 500": not a 500. `GetShopProductImages` returns `200 []` for an unknown product (`getImagesFromAllShop` / `getAssociatedShopIds` never throw `ProductNotFoundException`). So there is no unhandled exception, and CI is green.
- Shop-constraint and `ApiResourceMapping` mismatch flags are false positives too: this is an all-shops query by design, and the nested `{shopId, productImages:[{imageId, cover}]}` shape normalizes correctly (green happy path proves it).

**One real decision:** this endpoint behaves differently from its sibling `ProductImageList`, which maps `ProductNotFoundException → 404`. Either align (add `exceptionToStatus` + a 404 test) or keep `200 []` and drop the "unknown-id → 404" line from the description so the contract is accurate.

**Worth doing (test):** assert the nested `imageId`/`cover` values against a product with known images across shops, not just key presence + types, otherwise the per-shop grouping (the whole point vs the single-shop sibling) is exercised by nobody.

mattgoud:
Thanks for strengthening the test 👍 but the new fixture query broke CI on every integration leg:

```
SELECT `id_product` FROM `ps_image_shop` ORDER BY `id_product` ASC LIMIT 1 LIMIT 1
                                                                       ^^^^^^^^^^^^^
PrestaShopDatabaseException: SQL syntax ... near 'LIMIT 1'
```

The `LIMIT 1` is duplicated: the query string already ends with `LIMIT 1` and the DB helper (`getRow` / `getValue`) appends its own `LIMIT 1`. Drop the explicit `LIMIT 1` from the SQL (or use the helper without it) and the legs go green.

PrestaEdit:
Good catch, thanks @mattgoud! Fixed in d352d14 — dropped the explicit `LIMIT 1` since `Db::getValue()` already appends its own.

mattgoud (approved):
LGTM ✅ verified the wiring against core: `GetShopProductImages` returns the per-shop `{imageId, cover}` grouping, unknown product → empty `200 []` (no exception, so no `exceptionToStatus` needed), and it's intentionally all-shops so no `shopConstraint` mapping. the test asserts the full contract against the `image_shop` fixture.

non-blocking nit: if the test fixture has more than one shop, a quick assertion that the grouping actually spans multiple shops would nail down the endpoint's whole point — but the grouping logic itself lives in core, so fine to merge as-is.

tleon (approved):

jolelievre:

Note: in #410 the endpoint returns a single resource shared with the write operation: {productId, shopImages: [{shopId, images: [{imageId, cover}]}]}.

PR #361, opened 2026-07-13
Add GetProductStockMovements Admin API endpoint

PrestaEdit:
| Description?      | Adds `GET /products/{productId}/stock-movements` using `CQRSGetCollection` with `GetProductStockMovements`. Items expose `type`, `edition`, `fromOrders`, `stockMovementIds[]`, `stockIds[]`, `orderIds[]`, `employeeIds[]`, `employeeName`, `deltaQuantity`, `dates{}`. Uses `[_context][shopId]` mapping to auto-fill the required shopId from the API client context; offset/limit rely on the query's ctor defaults (0, 5). |
| How to test?      | Run `composer phpunit-integration`; new `ProductStockMovementsEndpointTest` covers scope protection and the list happy path. |

### DIFF
--- a/src/ApiPlatform/Resources/Product/ProductStockMovements.php
+++ b/src/ApiPlatform/Resources/Product/ProductStockMovements.php
@@ -24,6 +24,8 @@ namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\Product;
 
 use ApiPlatform\Metadata\ApiProperty;
 use ApiPlatform\Metadata\ApiResource;
+use ApiPlatform\Metadata\Parameters;
+use ApiPlatform\Metadata\QueryParameter;
 use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductNotFoundException;
 use PrestaShop\PrestaShop\Core\Domain\Product\Stock\Exception\StockAvailableNotFoundException;
 use PrestaShop\PrestaShop\Core\Domain\Product\Stock\Query\GetProductStockMovements;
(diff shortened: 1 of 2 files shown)


mattgoud:

**False positive to discard:** the "`shopConstraint` vs `shopId` (potential blocker)" point. `GetProductStockMovements::__construct(int $productId, int $shopId, int $offset = 0, int $limit = 5)` takes a plain `int $shopId`, so `'[_context][shopId]' => '[shopId]'` is exactly right; switching to `shopConstraint` would break it. CI is green.

**Worth doing (tests):**
- The happy path only asserts `type`/`deltaQuantity` inside the loop, and that loop may never run if the chosen product has no movements. Assert all fields against a product with known stock history.
- Add a 404 test for an unknown `productId` to actually exercise the `ProductNotFoundException → 404` mapping.

**Follow-up (not blocking):** `offset`/`limit` fall back to the ctor defaults (0, 5), so the endpoint is silently capped at 5 movements with no way to page. Worth exposing them as query parameters, or documenting the cap.

mattgoud:
The added stock-movement seeding fails on every integration leg with a core-side TypeError:

```
TypeError: PrestaShopBundle\Entity\StockMvt::setIdEmployee(): Argument #1 ($idEmployee) must be of type int, null given,
called in src/Core/Stock/StockManager.php on line 392
```

The seed creates a movement while no employee is set in the context, and `StockManager` doesn't guard against a null employee id. So this isn't a test-assertion issue — the seeding path hits an unguarded core code path. Two options:
- set a context employee before seeding the movement (e.g. via the API client / test context employee) so `id_employee` is non-null, or
- if the null-employee path should be supported, it needs a core guard (same family as PrestaShop/PrestaShop#41803).

The first is the quick unblock for this PR.

PrestaEdit:
Cross-ref: the null-employee core path this test was hitting is fixed in PrestaShop/PrestaShop#41803 (currently *Waiting for QA*). This PR sidesteps it by pinning `Employee(1)` in the context before seeding; once #41803 lands, the guard becomes unnecessary but harmless.

mattgoud (changes requested):
requesting changes on one point 🙏

the endpoint is capped at `GetProductStockMovements::DEFAULT_LIMIT` (5) and doesn't expose `offset`/`limit`, so a caller can only ever get the 5 most recent movements with no way to page through the rest. for a stock-movements listing that's a real limitation.

could you wire `offset`/`limit` as query params onto the query ctor (it already takes `int $offset = 0, int $limit = 5`)? or if paging is intentionally out of scope for now, document the cap explicitly in the description.

rest looks good btw — checked the `[_context][shopId] => [shopId]` mapping against the core ctor (plain int, so it's correct), the DTO matches the `StockMovement` result, and the 404 + full-field assertions are all there. just the pagination.

PrestaEdit:
Thanks for the review @mattgoud! Addressed in b3f1c8e…HEAD:

- Declared `offset` and `limit` as `QueryParameter`s on the operation (and mirrored them in `openapiContext` for the OpenAPI schema).
- They flow through `QueryProvider`'s context filters merge into the `GetProductStockMovements` ctor by parameter name, keeping the `DEFAULT_LIMIT = 5` fallback when omitted.
- Added `testListProductStockMovementsRespectsOffsetAndLimit` — seeds extra movements, asserts `limit=2` caps the response and `offset=1` shifts the window.

mattgoud:
thanks @PrestaEdit, pagination approach + the offset/limit test look right 👍

heads up though — `36cf1c3` is red on phpstan (all 4 versions):

```
ProductStockMovements.php:47 & :53 — Parameter $schema of QueryParameter constructor
expects array{type?: string, default?: string}|null,
array{type: 'integer', minimum: 0, default: 0|5} given.
```

the `QueryParameter` `schema` only accepts `{type?: string, default?: string}`, so `minimum` (extra key) and the integer `default` trip it on both params. simplest fix: keep the rich schema (`minimum` + numeric `default`) only in `openapiContext` (you already duplicated it there) and reduce the `QueryParameter` schema to `['type' => 'integer']`, or drop the `schema` arg entirely.

once phpstan's green i'll re-review 👍

mattgoud (approved):
pagination sorted and CI's green now — thanks @PrestaEdit 🙏

`offset`/`limit` exposed as query params (with the `DEFAULT_LIMIT` fallback), phpstan happy after the schema simplification, and `testListProductStockMovementsRespectsOffsetAndLimit` covers the cap + window. LGTM ✅

tleon (approved):

jolelievre:

Note: the stock-movements fixture is now fully API-based in #410 thanks to the new PUT /products/{productId}/stock endpoint (no more Employee(1) context workaround — the endpoint depends on PrestaShop/PrestaShop#41803 instead).

PR #363, opened 2026-07-13
Add SendEmployeePasswordResetEmail Admin API endpoint

PrestaEdit:
| Questions         | Answers                                                     |
|-------------------|-------------------------------------------------------------|
| Branch?           | dev                                                          |
| Description?      | Adds `POST /employees/password-resets` using `CQRSCreate` with `SendEmployeePasswordResetEmailCommand`. Body: `{email}`. |
| Type?             | new feature                                                  |
| Category?         | CO                                                           |
| BC breaks?        | no                                                           |
| Deprecations?     | no                                                           |
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630                       |
| How to test?      | Run `composer phpunit-integration`; new `EmployeePasswordResetEndpointTest` seeds the shop admin email, POSTs, verifies 201. Safe in CI: `PS_MAIL_METHOD = METHOD_DISABLE (3)` in the test env so `Mail::send` returns true without contacting an MTA (`classes/Mail.php:220`). |

First endpoint in the `Employee/` folder — no in-progress PR touches this area.

PR #364, opened 2026-07-13
Add ResetEmployeePassword Admin API endpoint

PrestaEdit:
| Description?      | Adds `POST /employees/password-reset-confirmations` using `CQRSCreate` with `ResetEmployeePasswordCommand`. Body: `{resetToken, password}`. Consumes the reset token issued by `SendEmployeePasswordResetEmailCommand` (#363). Invalid or expired tokens produce `InvalidResetPasswordTokenException` → HTTP 422. |
| How to test?      | Run `composer phpunit-integration`; new `EmployeePasswordResetConfirmationEndpointTest` seeds a 40-char token + future validity via `\Db::update()`, POSTs, and verifies the `passwd` column changed. |


tleon:
Hello you have some tests failing could you have a look ? maybe start with a rebase ? Thx in advance.

PrestaEdit:
Thanks @tleon — the branch was already at the tip of `dev`, so no rebase needed. The failures traced to a wrong import: `InvalidResetPasswordTokenException` is under `PrestaShopBundle\Security\Admin\Exception`, not `Core\Domain\Employee\Exception` (that one doesn't exist). PHPStan therefore failed on every core matrix, and the invalid-token test returned 500 instead of 422 because `exceptionToStatus` couldn't match the real exception. Fixed in the last push.

src/ApiPlatform/Resources/Employee/ResetEmployeePassword.php
```diff
@@ -0,0 +1,52 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Employee\Command\ResetEmployeePasswordCommand;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSCreate;
+use PrestaShopBundle\Security\Admin\Exception\InvalidResetPasswordTokenException;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Validator\Constraints as Assert;
+
+#[ApiResource(
+    operations: [
+        new CQRSCreate(
+            uriTemplate: '/employees/password-reset-confirmations',
```
nicosomb, line 35: why `confirmations` and why not `confirmation`?

PrestaEdit: Following the REST convention already used everywhere else in this module — every collection URI is plural (`/employees`, `/customers/groups`, `/attributes/groups`, `/features/values`, `/tax-rules-groups`…) and a `POST` on the collection creates one item. So `POST /employees/password-reset-confirmations` creates a password-reset confirmation, consistent with the rest of the resources. Happy to switch to singular if you'd rather standardise differently — but it would be the only singular collection in the module.

jolelievre:
```suggestion
            uriTemplate: '/employees/reset-password',
```
Hmm, but it's not a collection here, it's an action on an employee to reset their password So we keep `/employees` to respect the plural form, and the latter part is the action name It's not new it's already the convention we use in other endpoints as well

I'm not sure to understand the confirmations part you're talking about, especially since the command doesn't return anything

PrestaEdit: Good point on both counts — pushed the rename.

- URI is now `/employees/reset-password` to match the action-suffix convention (`/modules/upload-source`, `/categories/bulk-update-status`, etc.).
- Dropped the "confirmation" framing entirely: renamed the resource to `ResetEmployeePassword` and the test to `ResetEmployeePasswordEndpointTest`. You're right that since the command returns nothing, treating it as an action on `/employees` is cleaner than modeling a fake sub-resource.

jolelievre (changes requested):
After thjinking it through I don't think this one is useful, it makes sense in the BO because it is used when the employee wlicks on the email link and reset their password, but it only makes sense in this BO workflow:
- ask for reset pasword email email
- click on the link in the email that contains the reset token
- change the password (this step)

Although I can really imagine a use case where we handle the email sending via the API, which is handled in this https://github.com/PrestaShop/ps_apiresources/pull/363 and makes sense

But the last part of the workflow needs to be handled by the employee via the email link and the BO, to use it with the API you'd need to copy the reset token from the mail (so only employee received it) and then pass it to the Admin API which is used server to server, I don't think it makes sense

We could have a usage for an admpin API that simply allows resetting the password ass is but that would be another endpoint (and I think the employee edition already handles this)

PrestaEdit:
Fair point, and after checking I agree with closing this one.

I verified that `EditEmployeeCommand` already accepts a `plainPassword` field, so the "admin resets another employee's password" use case will be covered naturally once an `Employee` resource is exposed via `CQRSUpdate` — no reset token needed, which is the right shape for a server-to-server Admin API.

For the token-based flow, you're right that it doesn't fit: nobody is going to relay a token received by email into an Admin API call. #363 (`SendEmployeePasswordResetEmail`) still makes sense on its own since it triggers the native BO workflow, but the confirmation step belongs in the BO.

Closing this PR. Happy to open a follow-up for a full `Employee` resource (Add/Edit/Delete) if that's on the roadmap — the password change would come along for free with the Edit endpoint.

PR #366, opened 2026-07-13
Add QuickAccess Admin API endpoints (CRUD + bulk-delete + toggle)

PrestaEdit:
| Description?      | Full CRUD for the QuickAccess admin bookmarks: create / read / edit / delete + bulk-delete, plus a per-item toggle of the `new_window` column via `PUT /quick-accesses/{id}/new-window-toggles`. Uses the classic ID-only-ctor + setters `EditCommand` pattern, and standard `BulkDelete*` layout in a companion resource file. |
| How to test?      | Run `composer phpunit-integration`; `QuickAccessEndpointTest` walks the full CRUD chain via `@depends` (create → get → edit → toggle → delete), then a separate bulk-delete test seeds two rows via the endpoint and verifies row-count = 0. |

**⚠️ Depends on PR #220** — the `QuickAccess` core CQRS classes were introduced in PS 9.1+/develop and do not exist at the 9.0.3 tag. The 9.0.3 integration / PHPStan CI legs will fail here until #220 (drop 9.0.3 from the module CI matrix) lands. The endpoints are functional on 9.1.x and develop.

### DIFF
--- a/src/ApiPlatform/Resources/QuickAccess/BulkQuickAccesses.php
+++ b/src/ApiPlatform/Resources/QuickAccess/BulkQuickAccesses.php
@@ -1,0 +1,55 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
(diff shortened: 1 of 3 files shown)


PrestaEdit:
Closing: the QuickAccess commands are declared out of Admin API scope on `dev`.

`src/Command/GenerateApiTrackingTableCommand.php` lists `AddQuickAccessCommand`, `EditQuickAccessCommand`, `DeleteQuickAccessCommand`, `BulkDeleteQuickAccessCommand`, `ToggleQuickAccessNewWindowCommand` and `GetQuickAccessForEditing` in `EXCLUDED_CQRS_CLASSES` with the reason *"Quick access links are a back office UI customization feature"* — which is exactly the set of endpoints this PR adds.

Quick access links are per-employee back office shortcuts, so exposing them over the Admin API would contradict that decision. If it should be revisited, the place to do it is the exclusion list rather than a resource that bypasses it.

PR #374, opened 2026-07-13
Add SearchProductsForFreeGift Admin API endpoint

PrestaEdit:
| Description?      | Adds `GET /products/free-gift-candidates?phrase=[&limit=]` using `CQRSGetCollection` with `SearchProductsForFreeGift`. Uses `[_context][langId]` and `[_context][shopId]` mappings to auto-fill the two required scoping args from the API client context. Response items: `productId`, `name`, `reference`, `imagePath`, `productType`, `disabled`, `disabledReason`. |
| How to test?      | Run `composer phpunit-integration`; `FreeGiftCandidateEndpointTest` searches for `hum` against the demo catalog (Hummingbird products). |

**⚠️ Depends on PR #220** — `SearchProductsForFreeGift` was introduced in PS 9.1+/develop and does not exist at the 9.0.3 tag. The 9.0.3 CI matrix legs will fail here until #220 (drop 9.0.3 from CI) lands.

### DIFF
--- a/src/ApiPlatform/Resources/Product/FreeGiftCandidate.php
+++ b/src/ApiPlatform/Resources/Product/FreeGiftCandidate.php
@@ -1,0 +1,79 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+declare(strict_types=1);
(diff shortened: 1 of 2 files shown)


mattgoud (commented):
A couple of real defects here beyond the CI-matrix dependency:

1. **`disabledReason` must be nullable.** The core DTO getter is `ProductForFreeGift::getDisabledReason(): ?string` and the handler returns `null` for eligible products (`return [false, null];`). The resource declares `public string $disabledReason;` (non-nullable), so denormalizing an eligible product throws `The type of the "disabledReason" attribute must be "string", "NULL" given.` → **HTTP 500**. This is why the develop/9.2.x integration legs are red. Fix: `public ?string $disabledReason;`.

2. **`imagePath` doesn't map to the DTO.** The DTO exposes `getImageUrl()` (key `imageUrl`), not `imagePath`, and there's no `ApiResourceMapping`, so `imagePath` is never populated (currently masked by defect #1). Rename the property to `imageUrl`, or add `ApiResourceMapping ['[imagePath]' => '[imageUrl]']`.

3. **Dependency is under-stated.** `SearchProductsForFreeGift` was introduced in **9.2.x/develop only** (commit 7212d79) — it does **not** exist in 9.1.x. #220 drops only 9.0.3, so PHPStan(9.1.x) and the 9.1.x integration legs stay red after it merges. Either also drop 9.1.x from the matrix, or gate the resource on `class_exists`.

Nit: the happy-path test doesn't assert a non-empty result set and doesn't cover `imagePath`/`reference`/`disabledReason`, so it wouldn't have caught #1 or #2. Worth asserting at least one row with all declared fields.

jolelievre:

Note: #410 also fixes the nullable disabledReason (it caused a 500 for eligible products), exposes imageUrl and reference, and gates the operation with minVersion 9.2.0 since SearchProductsForFreeGift only exists since PrestaShop 9.2.

PR #377, opened 2026-07-13
Add CancelOrderProduct Admin API endpoint

PrestaEdit:
| Description?      | Adds `POST /orders/{orderId}/product-cancellations` using `CQRSCreate` with `CancelOrderProductCommand`. Body: `{cancelledProducts: {\"orderDetailId\": quantity}}` — the associative map is the Command's ctor arg verbatim. URI segment `product-cancellations` is idempotent under Rector (last word `cancellations` already plural). |
| How to test?      | Run `composer phpunit-integration`; scope-protection tests only — a full happy-path test would need to seed a paid, non-shipped order + a valid order-detail row set which the existing test infrastructure does not currently expose. |

### DIFF
--- a/src/ApiPlatform/Resources/Order/OrderProductCancellation.php
+++ b/src/ApiPlatform/Resources/Order/OrderProductCancellation.php
@@ -25,8 +25,8 @@ namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\Order;
 use ApiPlatform\Metadata\ApiProperty;
 use ApiPlatform\Metadata\ApiResource;
 use PrestaShop\PrestaShop\Core\Domain\Order\Command\CancelOrderProductCommand;
-use PrestaShop\PrestaShop\Core\Domain\Order\Exception\CannotCancelOrderProductException;
-use PrestaShop\PrestaShop\Core\Domain\Order\Exception\CartRuleValidityException;
+use PrestaShop\PrestaShop\Core\Domain\Order\Exception\InvalidCancelProductException;
+use PrestaShop\PrestaShop\Core\Domain\Order\Exception\InvalidOrderStateException;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\OrderException;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\OrderNotFoundException;
 use PrestaShopBundle\ApiPlatform\Metadata\CQRSCreate;
@@ -44,8 +44,8 @@ use Symfony\Component\Validator\Constraints as Assert;
     ],
     exceptionToStatus: [
         OrderNotFoundException::class => Response::HTTP_NOT_FOUND,
-        CannotCancelOrderProductException::class => Response::HTTP_UNPROCESSABLE_ENTITY,
-        CartRuleValidityException::class => Response::HTTP_UNPROCESSABLE_ENTITY,
+        InvalidCancelProductException::class => Response::HTTP_UNPROCESSABLE_ENTITY,
+        InvalidOrderStateException::class => Response::HTTP_UNPROCESSABLE_ENTITY,
         OrderException::class => Response::HTTP_UNPROCESSABLE_ENTITY,
     ],
 )]
--- a/src/ApiPlatform/Resources/Order/OrderProductCancellation.php
+++ b/src/ApiPlatform/Resources/Order/OrderProductCancellation.php
@@ -1,0 +1,67 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\Order;
+
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ApiResource;
+use PrestaShop\PrestaShop\Core\Domain\Order\Command\CancelOrderProductCommand;
+use PrestaShop\PrestaShop\Core\Domain\Order\Exception\CannotCancelOrderProductException;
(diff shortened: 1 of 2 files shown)


mattgoud:
Two of the mapped exceptions don't resolve, which will break `composer phpstan` in CI:

- `PrestaShop\PrestaShop\Core\Domain\Order\Exception\CannotCancelOrderProductException` doesn't exist anywhere in core.
- `CartRuleValidityException` isn't in `Order\Exception` — it lives in `Core\Domain\CartRule\Exception`, so the import resolves to a phantom class.

`CancelOrderProductHandler` only throws `InvalidCancelProductException` and `InvalidOrderStateException`, both of which extend `OrderException` and are therefore already covered by your `OrderException => 422` catch-all. So both entries can just be dropped (optionally map `InvalidCancelProductException` explicitly).

Also, the test only implements `getProtectedEndpoints()` — no happy-path and no `testInvalid*` (422) test, which the module testing conventions require.

mattgoud (changes requested):
Requesting changes: the mapped `CannotCancelOrderProductException` doesn't exist in core and `CartRuleValidityException` is imported from the wrong namespace, so `composer phpstan` will fail. Details in my comment above.

PrestaEdit:
Thanks @mattgoud — addressed in the last push. Both bad references removed:

- `CannotCancelOrderProductException` doesn't exist in core; `CancelOrderProductHandler` actually throws `InvalidCancelProductException`. Swapped.
- `CartRuleValidityException` isn't in the Order namespace (it lives under `Core\Domain\CartRule\Exception`), and the handler doesn't throw it — removed.
- Added `InvalidOrderStateException` while I was in there, which the handler does throw when cancelling a product in an incompatible order state.

Unblocks `composer phpstan` in CI.

Ready for another look.

mattgoud:
Thanks @PrestaEdit, the phantom exception imports are gone and CI is green now 👍 The only thing still open from my review is the integration test: the test only implements `getProtectedEndpoints()`. Could you add a happy-path plus a `testInvalid*` (422) case (e.g. an invalid product/order making the cancel fail) before it goes for final review? Thanks!

PrestaEdit:
Thanks @mattgoud! Added three live 422 tests + a skipped happy-path scaffold (a08b… on top).

- `testEmptyCancelledProductsReturns422` — `Assert\NotBlank` validation.
- `testInvalidQuantityReturns422` — `InvalidCancelProductException::INVALID_QUANTITY` (quantity 0). `checkInput()` runs before `checkOrderState()`, so this fires even on a paid fixture order.
- `testCancelOnAlreadyPaidOrderReturns422` — `InvalidOrderStateException::ALREADY_PAID` on a fixture order → exercises the full handler chain end-to-end.

For the actual happy-path I hit the same trap as #271: a successful cancel calls `OrderProductQuantityUpdater::reinjectQuantity` → `StockAvailable::updateQuantity` → `Core\Stock\StockManager::saveMovement()`, which does `$stockMvt->setIdEmployee($employee->id)` on the null-id API-context employee → `TypeError`. The guard is in core PR PrestaShop/PrestaShop#41803, still open.

I've added a `testCancelProductHappyPath` gated with `markTestSkipped` referencing #41803 so it's ready to enable (with unpaid-order + order_detail seeding) once that fix ships, mirroring what #271 does. Happy to convert this PR to draft the same way if you'd prefer — let me know.

mattgoud:
The re-run failure isn't flaky, it's a deterministic bug in the test's `setUpBeforeClass()`. The query passed to `Db::getRow()` ends with an explicit `LIMIT 1`, but `getRow()` appends its own ` LIMIT 1` (see `classes/db/Db.php`: `$sql = rtrim($sql, " \t\n\r\0\x0B;") . ' LIMIT 1';`). The final SQL becomes `... LIMIT 1 LIMIT 1`, which throws:

```
PrestaShopDatabaseException: You have an error in your SQL syntax ... near 'LIMIT 1' at line 4
SELECT o.`id_order`, od.`id_order_detail` ...
OrderProductCancellationEndpointTest.php:43
```

Fix: drop the explicit `LIMIT 1` from the query and let `getRow()` add it:

```php
$row = \Db::getInstance()->getRow(
    'SELECT o.`id_order`, od.`id_order_detail`
     FROM `' . _DB_PREFIX_ . 'orders` o
     INNER JOIN `' . _DB_PREFIX_ . 'order_detail` od ON od.`id_order` = o.`id_order`'
);
```

Two notes:
- The failing leg changes between runs because CI shards the test classes across matrix jobs, so this class lands on a different job each time. The failure itself is systematic, not intermittent.
- The two `OrderReturnEndpointTest::testProtectedEndpoints` failures (expected 200) in the same run are most likely collateral: the aborted `setUpBeforeClass` leaves the DB in an unclean state. They should clear once the `LIMIT 1` is fixed. Worth re-running to confirm.

mattgoud:
The `LIMIT 1` fix is in and working (the SQL syntax error is gone), thanks. But it uncovered a deeper blocker on all 8 integration legs:

```
testCancelOnAlreadyPaidOrderReturns422
TypeError: PrestaShopBundle\Entity\StockMvt::setIdEmployee(): Argument #1 ($idEmployee) must be of type int, null given
  StockManager.php:392  ←  CancelOrderProductHandler.php:185  ←  ...EndpointTest.php:87
```

This is the same root cause as the skipped happy-path: in the API context there is no logged-in employee, so the cancel path's stock movement calls `StockMvt::setIdEmployee(null)` and 500s. It's exactly core #41803 ("Guard stock movement when the context employee has no id").

Two things here:
- The negative test picks "any fixture order" via `LIMIT 1`, but that order isn't necessarily in a state that rejects early, so instead of a clean 422 it proceeds into the cancel and hits the stock-movement bug. It needs a fixture order that is genuinely in the already-paid/invalid state so the `InvalidOrderStateException` fires before any stock movement.
- Even with that, the endpoint's real cancel path stays blocked until core #41803 lands (no employee id in the API context). So this PR is effectively gated on #41803, same as its happy-path.

PrestaEdit:
Thanks @mattgoud — pushed the fixture fix.

The setUp query now filters on `INNER JOIN order_payment op ON op.order_reference = o.reference`, so `Order::hasPayments()` is guaranteed true on the selected fixture. That makes `checkOrderState()` short-circuit to `InvalidOrderStateException::ALREADY_PAID` (422) before the handler ever reaches the stock-movement path, so the negative test no longer surfaces the `StockMvt::setIdEmployee(null)` TypeError.

Agreed on the deeper gate: the real happy-path stays skipped until PrestaShop/PrestaShop#41803 lands. That PR is open and marked mergeable — once it ships in the tested core versions I'll drop the skip and add the unpaid-order seeding for a full success-path leg.

PrestaEdit:
@mattgoud friendly ping — everything you raised is in:
- Phantom exception imports (`CannotCancelOrderProductException`, `CartRuleValidityException`) → removed, real handler throws (`InvalidCancelProductException`, `InvalidOrderStateException`) mapped.
- Integration tests: 3 live 422 tests + skipped happy-path scaffold.
- `LIMIT 1` double-suffix bug → fixed.
- Fixture query now filters on `order_payment` so `checkOrderState()` short-circuits to `ALREADY_PAID` (422) before the stock-movement path, no more `StockMvt::setIdEmployee(null)` collateral in the negative test.

Last push 97f63e8 on 2026-07-15. The happy-path stays gated on core #41803 (retargeted to `9.2.x` today — awaiting core merge). Would appreciate a re-review to lift the stale `CHANGES_REQUESTED`.

PR #378, opened 2026-07-13
Add AddProductToOrder Admin API endpoint

PrestaEdit:
| Description?      | Adds `POST /orders/{orderId}/product-additions` using `CQRSCreate` with `AddProductToOrderCommand`. Body: `{productId, combinationId?, productPriceTaxIncluded, productPriceTaxExcluded, productQuantity, orderInvoiceId?, hasFreeShipping?}`. |
| How to test?      | Run `composer phpunit-integration`; scope-protection test only for now. |

**⚠️ Depends on PrestaShop/PrestaShop#42045** — that PR makes `AddProductToOrderCommand::__construct` public so Symfony's serializer can instantiate it from the request body. Until #42045 lands, the resource is registered but any actual POST will fail at denormalization (private ctor). The CI legs that exercise the endpoint will go green automatically as soon as the core change reaches the module CI matrix — no code change needed here.


mattgoud:
`orderInvoiceId` and `hasFreeShipping` won't actually be applied. `AddProductToOrderCommand` (in `Core\Domain\Order\Product\Command`) has a private constructor and sets those two values only through the named constructors `withNewInvoice()` / `toExistingInvoice()` — they are not constructor parameters and have no setters. Since `CommandProcessor` builds the command via its constructor, both fields are silently dropped: free shipping can't be set and the "add to an existing invoice" path is unreachable (a new invoice is always created).

Note this endpoint also depends on core #42045 (constructor visibility), but that change alone won't fix the two dropped fields.

Test-wise: only `getProtectedEndpoints()` is present, no happy-path / `testInvalid*` (422).

mattgoud (changes requested):
Requesting changes: `orderInvoiceId` and `hasFreeShipping` are set only via the named constructors and are silently dropped by `CommandProcessor` (which builds via the constructor). Details in my comment above.

PrestaEdit:
Thanks @mattgoud — addressed in the last push. Went with a dedicated processor since a config-only fix can't route through the named constructors:

- Added `src/ApiPlatform/Processor/AddProductToOrderProcessor.php` implementing `ApiPlatform\State\ProcessorInterface`. It inspects the DTO and calls:
  - `AddProductToOrderCommand::toExistingInvoice(...)` when `orderInvoiceId` is a positive int, or
  - `AddProductToOrderCommand::withNewInvoice(..., $hasFreeShipping)` otherwise.
- Registered the processor as `public: true` under `config/admin/services.yml`.
- Wired it on the operation via `processor: AddProductToOrderProcessor::class`.

Both `orderInvoiceId` and `hasFreeShipping` are now actually applied. Ready for another look.

mattgoud:
Routing through `AddProductToOrderProcessor` does make `orderInvoiceId` / `hasFreeShipping` reach the command, but it trips two CI checks:

1. PHPStan `ApiResourceProcessorProviderRule` forbids custom `ProcessorInterface` implementations in this module ("Use the CommandProcessor or QueryProvider from PrestaShopBundle instead").
2. PHP-CS-Fixer flags the new import ordering in `OrderProductAddition.php`.

Since those two values are only settable via the named constructors (`withNewInvoice()` / `toExistingInvoice()`), this really needs a core-side change (making them constructor params, or a mapping path the core `CommandProcessor` can use) rather than a module-local processor. Note core #42045 on its own won't cover this.

PrestaEdit:
Thanks — you're right that the module-local processor path is a dead end because of `ApiResourceProcessorProviderRule`. On the missing-fields point though, core #42045 does cover it: it promotes `orderInvoiceId` and `hasFreeShipping` to optional constructor params on `AddProductToOrderCommand` (both static factories `withNewInvoice()` / `toExistingInvoice()` are preserved as thin wrappers), so once it lands the core `CommandProcessor` can denormalize the request body straight into the command — no module-side processor required.

I've just pushed 7beb464 which:
- deletes `AddProductToOrderProcessor` and its service registration,
- drops `processor:` from the `CQRSCreate` operation,
- types the price fields as `string` (matches the JSON body and the command ctor).

The endpoint stays inert until #42045 reaches the module CI matrix, at which point the scope-protection test still passes and a real POST becomes functional.

tleon (approved):

PrestaEdit:
@mattgoud friendly ping — the module-side rework you flagged has been pushed (7beb464, on 2026-07-15): dropped the local processor, made the endpoint rely on core PR #42045 which promotes `orderInvoiceId` / `hasFreeShipping` to optional constructor params so the stock `CommandProcessor` picks them up. @tleon approved on 2026-07-29. Would appreciate a re-review to lift the stale `CHANGES_REQUESTED`. (Note: this PR still awaits core #42045 to actually run, but that's independent.)

PR #379, opened 2026-07-13
Add GetOrderProductsForViewing Admin API endpoint

PrestaEdit:
| Description?      | Adds `GET /orders/{orderId}/products?offset=&limit=&productsSorting=` using `CQRSGet` with `GetOrderProductsForViewing`. The response is a single `OrderProducts` resource: `{orderId, products: [...]}`. Each row exposes per-line refund state (`amountRefundable`, `quantityRefundable`, `refundable`, …), quantities and prices. Complements #298 (order details read). |
| How to test?      | Run `composer phpunit-integration`; `OrderProductsEndpointTest` fetches the first-order's products and asserts each row has `productId`/`name`/`quantity`. |

**⚠️ Depends on PrestaShop/PrestaShop#42045** — `GetOrderProductsForViewing` had no explicit ctor at all (only `new self()` + static factories filling fields). #42045 adds a public `__construct(int $orderId = 0, ?int $offset, ?int $limit, string $productsSorting)`. Until it lands the endpoint 500s at denormalization; it will turn green automatically after the core change reaches this module's CI dependencies.


mattgoud:
The `CQRSGetCollection` shape doesn't match the query. `GetOrderProductsForViewingHandler::handle()` returns a single `OrderProductsForViewing` wrapper object (the rows are behind `getProducts()`), whereas `CQRSGetCollection` expects the handler to return a plain list of items (cf. `GetProductImagesHandler::handle(): array`). So the collection provider receives a wrapper, not the array of product rows the DTO/test assume — the response won't be what's expected.

Minor: `public int $id;` goes against the "never use `id` as a property name" convention.

mattgoud (changes requested):
Requesting changes: `CQRSGetCollection` expects a list, but `GetOrderProductsForViewingHandler::handle()` returns the `OrderProductsForViewing` wrapper, so the response shape won't match. Details in my comment above.

PrestaEdit:
Thanks @mattgoud — addressed in the last push. Wrote a dedicated provider since the wrapper can't be unwrapped through configuration alone:

- Added `src/ApiPlatform/Provider/OrderProductListProvider.php` implementing `ApiPlatform\State\ProviderInterface`. It dispatches `GetOrderProductsForViewing::all(...)`, calls `->getProducts()` on the returned wrapper, and paginates the resulting array in memory (the wrapper exposes no total count, so `::paginated()` alone can't drive `PaginationElements`).
- Wired it on the operation via `provider: OrderProductListProvider::class` and registered the service `public: true` in `config/admin/services.yml`.
- Renamed `$id` → `$productId` on the resource (module convention), bridged via `ApiResourceMapping`. Updated the endpoint test to assert `productId`.

Ready for another look.

mattgoud:
Same blocker as #378 on the provider side: `OrderProductListProvider implements ProviderInterface` is rejected by the module's PHPStan `ApiResourceProcessorProviderRule` (custom providers aren't allowed, use the core `QueryProvider`). On top of that the happy-path test still errors:

```
assertArrayHasKey(): Argument #2 ($array) must be of type array, int given
```

so the response isn't the list of product rows yet. Root cause is that `GetOrderProductsForViewingHandler` returns the `OrderProductsForViewing` wrapper rather than a plain list. Rather than a custom provider, this is probably best solved either in core (a query/handler that returns the list directly) or by modeling it as a `CQRSGet` returning a single object with a `products` array property. Happy to help figure out the cleanest path.

PrestaEdit:
Thanks — good call. I went with the second option: dropped `OrderProductListProvider` entirely, reshaped the resource as a single `OrderProducts` (`CQRSGet`) with a `products` array property, and kept the `id -> productId` rename via an `ApiResourceMapping` `[@index]` wildcard on the nested rows. The core `QueryProvider` now handles the dispatch and normalizes `OrderProductsForViewing` directly, so the PHPStan rule passes and the wrapper's shape drives the response. Test updated for the new envelope (`{orderId, products: [...]}`). Still depends on PrestaShop/PrestaShop#42045 for the query's public ctor.

PrestaEdit:
@mattgoud friendly ping — the reshape you suggested is in bddbbd7 (2026-07-15):
- Dropped `OrderProductListProvider` (PHPStan-blocked).
- Reshaped resource as single `OrderProducts` (`CQRSGet`) with a `products` array property; core `QueryProvider` handles the dispatch and normalizes `OrderProductsForViewing` directly.
- `id → productId` rename via `ApiResourceMapping` `[@index]` wildcard.
- Test updated for the new envelope `{orderId, products: [...]}`.

Still depends on core #42045 (public constructor on `GetOrderProductsForViewing`) — that PR is still open on 9.2.x. Once it merges, the endpoint runs end-to-end. No changes needed on this PR right now, just a re-review would help lift the stale `CHANGES_REQUESTED`.

PR #380, opened 2026-07-13
Add CmsPageCategory localized-name Admin API endpoint

PrestaEdit:
| Description?      | Adds `GET /cms-page-categories/{cmsPageCategoryId}/localized-name` using `CQRSGet` with `GetCmsPageCategoryNameForListing`. Uses `[_queryResult]` scalar wrap — the handler returns a plain string (the localized category name in the context language) surfaced as `{name}`. |
| How to test?      | Run `composer phpunit-integration`; `CmsPageCategoryLocalizedNameEndpointTest` fetches id=1 (ROOT) and expects a string name. |

**⚠️ Depends on PrestaShop/PrestaShop#42046** — that PR adds an explicit ctor arg (`?int $cmsPageCategoryId`) to the query. Without it the handler reads `id_cms_category` from the current HTTP request instead, which means the URL segment `{cmsPageCategoryId}` is ignored and the response is always the root-category name. The test targets id=1 = ROOT so it stays green during the transition; it will validate the URL-driven id once the core change reaches this module's CI dependencies.

### DIFF
--- a/src/ApiPlatform/Resources/CmsPageCategory/CmsPageCategoryLocalizedName.php
+++ b/src/ApiPlatform/Resources/CmsPageCategory/CmsPageCategoryLocalizedName.php
@@ -30,7 +30,7 @@ use PrestaShopBundle\ApiPlatform\Metadata\CQRSGet;
 #[ApiResource(
     operations: [
         new CQRSGet(
-            uriTemplate: '/cms-page-categories/{cmsPageCategoryId}/localized-name',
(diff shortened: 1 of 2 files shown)


tleon (approved):

PrestaEdit:
Superseded by #433, which merges this endpoint with the rest of the domain, following the mutualisation asked for by the core team (one PR per domain rather than one per endpoint).

PR #383, opened 2026-07-13
Add VirtualProductFile Admin API endpoints (add + update)

PrestaEdit:
| Description?      | Adds:<br/>- `POST /products/{productId}/virtual-file` (AddVirtualProductFileCommand — body `{filePath, displayName, accessDays?, downloadTimesLimit?, expirationDate?}`)<br/>- `PATCH /virtual-product-files/{virtualProductFileId}` (UpdateVirtualProductFileCommand — ID-only ctor + optional setters).<br/>`filePath` in the POST body points to a temp file on disk that the uploader will validate + copy to `_PS_DOWNLOAD_DIR_`. |
| How to test?      | Scope-protection tests only — happy-path coverage requires seeding a TYPE_VIRTUAL product AND placing a real file at the target `filePath`, which the existing test infrastructure does not currently expose. |

### DIFF
--- a/src/ApiPlatform/Resources/Product/VirtualProductFile.php
+++ b/src/ApiPlatform/Resources/Product/VirtualProductFile.php
@@ -1,0 +1,79 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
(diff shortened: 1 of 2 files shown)


mattgoud (commented):
Thanks for this 👍 One blocker before merge:

**`Rector Dry Run` is red** on the URI templates (`ApiResourceUriTemplateRector`). The rule is right about the POST route but its raw suggestion for the PATCH route is broken (`/products/{virtualProductFileId}` collides with Product's own `/products/{productId}`). Please align both to the module's own-id sub-resource convention (cf. `ProductImage`, `Combination`):

- POST → `/products/{productId}/virtual-files`
- PATCH → `/products/virtual-files/{virtualProductFileId}`

Both survive the Rector rule unchanged, so the gate goes green.

Then update the protected-endpoint paths in `VirtualProductFileEndpointTest.php` (~L33-34) to match (`/products/1/virtual-files` and `/products/virtual-files/1`), otherwise the scope test asserts on routes that no longer exist.

Nit (optional): `exceptionToStatus` doesn't map `CannotAddVirtualProductFileException` / `CannotUpdateVirtualProductFileException`, so those fall through to 500. Might be worth mapping explicitly.

CQRS wiring itself is correct and there's no core dependency here.

tleon (approved):

jolelievre:

Note: in #410 the POST returns the created file (through GetProductForEditing), the Cannot* exceptions are mapped to 422, and the URIs are singularized to /products/{productId}/virtual-file and /products/virtual-file/{virtualProductFileId} since a product has at most one virtual file.

PR #384, opened 2026-07-13
Add SetProductImagesForAllShop Admin API endpoint

PrestaEdit:
| Description?      | Adds `PUT /products/{productId}/image-shop-associations` using `CQRSUpdate` with `SetProductImagesForAllShopCommand`. Body: `{productImageSettings: [{productImageId, shopIds[]}, ...]}` — mirrors the `ProductImageSetting` VO shape. Each entry maps an image to a set of shops the image should belong to; images not listed are detached from all shops. |
| How to test?      | Scope-protection test only for now — full happy-path coverage would need multi-shop fixtures + seeded product images that the existing test infrastructure does not currently expose. |

**⚠️ Depends on PrestaShop/PrestaShop#42049** — that PR adds a public `setProductImageSettings(array)` bulk setter to the command. Today only the per-item `addProductSetting()` adder exists, which Symfony's serializer can't drive from a JSON body. After #42049 lands the endpoint will start denormalizing correctly.

### DIFF
--- a/src/ApiPlatform/Resources/Product/ProductShopImagesAssociation.php
+++ b/src/ApiPlatform/Resources/Product/ProductShopImagesAssociation.php
@@ -1,0 +1,67 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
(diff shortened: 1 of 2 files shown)


tleon (approved):

jolelievre:

Note: in #410 the operation becomes PUT /products/{productId}/shop-images, sharing its URI and resource with the read side, and it no longer depends on PrestaShop/PrestaShop#42049: the command is built by a dedicated module denormalizer, so the endpoint works on every supported core version.

PR #385, opened 2026-07-13
Add Order refund + product-return Admin API endpoints

PrestaEdit:
| Description?      | Adds:<br/>- `POST /orders/{orderId}/refunds` (IssueStandardRefundCommand)<br/>- `POST /orders/{orderId}/product-returns` (IssueReturnProductCommand)<br/>Body shape for both: `{orderDetailRefunds: [{orderDetailId, productQuantity, refundedAmount?}, ...], refundShippingCost, generateCreditSlip, generateVoucher, voucherRefundType}` — plus `restockRefundedProducts` on the product-return endpoint. `refundedAmount` is optional per line (partial refund vs full-line refund). |
| How to test?      | Scope-protection tests only — full happy-path coverage would need seeding an order with paid+shipped order-detail rows, which the existing test infrastructure does not currently expose. |

**⚠️ Depends on PrestaShop/PrestaShop#42050** — that PR exposes `OrderDetailRefund::__construct` as public. Both commands take `array $orderDetailRefunds` expecting `OrderDetailRefund[]`; Symfony's serializer needs a usable public ctor to denormalize each entry from the JSON body.

### DIFF
--- a/src/ApiPlatform/Resources/Order/OrderProductReturn.php
+++ b/src/ApiPlatform/Resources/Order/OrderProductReturn.php
@@ -1,0 +1,82 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\Order;
+
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ApiResource;
+use PrestaShop\PrestaShop\Core\Domain\Order\Command\IssueReturnProductCommand;
+use PrestaShop\PrestaShop\Core\Domain\Order\Exception\OrderException;
+use PrestaShop\PrestaShop\Core\Domain\Order\Exception\OrderNotFoundException;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSCreate;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Validator\Constraints as Assert;
+
+#[ApiResource(
+    operations: [
+        new CQRSCreate(
+            uriTemplate: '/orders/{orderId}/product-returns',
+            requirements: ['orderId' => '\d+'],
+            CQRSCommand: IssueReturnProductCommand::class,
+            scopes: ['order_write'],
+        ),
+    ],
+    exceptionToStatus: [
+        OrderNotFoundException::class => Response::HTTP_NOT_FOUND,
+        OrderException::class => Response::HTTP_UNPROCESSABLE_ENTITY,
+    ],
+)]
+class OrderProductReturn
+{
+    public int $orderId;
+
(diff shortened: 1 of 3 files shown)


mattgoud:
The `orderDetailRefunds` shape is inverted on both resources. Both commands override `setOrderDetailRefunds()` to iterate an **associative array keyed by orderDetailId** with a single `quantity` field:

```php
foreach ($orderDetailRefunds as $orderDetailId => $detailRefund) {
    $this->orderDetailRefunds[] = OrderDetailRefund::createStandardRefund(
        $orderDetailId,
        $detailRefund['quantity']
    );
}
```

Expected body: `{"orderDetailRefunds": {"<orderDetailId>": {"quantity": N}}}`.

The DTOs instead document a **list** `[{orderDetailId, productQuantity, refundedAmount?}]`. Passed as-is, the loop reads numeric keys (0, 1, …) as order-detail IDs and `$detailRefund['quantity']` is an undefined index (only `productQuantity` is provided) — broken at runtime.

`refundedAmount` is also dead: both commands only ever call `createStandardRefund` (quantity-only, full-line), so partial per-line amounts aren't supported here and shouldn't be exposed.

Test-wise: only `getProtectedEndpoints()` is present for both endpoints, no happy-path / `testInvalid*` (422).

mattgoud (changes requested):
Requesting changes: the `orderDetailRefunds` shape is inverted on both resources (the commands expect an assoc array keyed by orderDetailId with a `quantity` key, not a list). Details in my comment above.

PrestaEdit:
Thanks @mattgoud — addressed in the last push. Both resources now advertise the shape the commands actually consume:

```
orderDetailRefunds: {
  "12": { "quantity": 2 },
  "15": { "quantity": 1 }
}
```

- Flipped `OrderStandardRefund::$orderDetailRefunds` and `OrderProductReturn::$orderDetailRefunds` from a list-of-objects openapiContext to `type: object` + `additionalProperties: {type: object, properties: {quantity}, required: [quantity]}`, matching the same convention we used on [#320](https://github.com/PrestaShop/ps_apiresources/pull/320).
- Updated docblocks to note that this mirrors what `setOrderDetailRefunds()` iterates in each command.

Ready for another look.

mattgoud:
Thanks @PrestaEdit, the `orderDetailRefunds` shape is fixed on both resources and CI is green 👍 Still open from my review: the test only has `getProtectedEndpoints()` for both endpoints. Could you add a happy-path plus a `testInvalid*` (422) case for the standard-refund and product-return endpoints before final review? Thanks!

PrestaEdit:
Added `testInvalid*` (422) cases for both endpoints via missing `orderDetailRefunds` — pushed. Happy-path is the sticking point I flagged in the PR body: `IssueStandardRefund`/`IssueReturnProduct` require an order with paid + shipped order-detail rows, and the current `ApiTestCase` fixtures only give us the default install-dev order (no payment / no shippable state). Doing it properly means new fixture plumbing (seed a payment, force `OrderDetail` state) that felt out-of-scope for this PR. Happy to open a follow-up ticket for a shared "refundable order" fixture if you'd rather block on that here — let me know which you prefer.

mattgoud:
the `orderDetailRefunds` shape fix is spot on 👍 — updating my request: the resource + the two 422 tests are good, I'm only keeping changes requested for a **happy-path** test (issue an actual refund and assert the effect — credit slip / voucher / restocked qty), not just the missing-refunds 422 path. i know seeding a refundable order is heavy, so no rush; just flagging it's the one thing left before approval.

PrestaEdit:
Thanks @mattgoud — happy-path added in bf85dd3.

The new `testIssueStandardRefundGeneratesCreditSlip` promotes the first seeded order to `PS_OS_PAYMENT` in `setUpBeforeClass` (which also generates the invoice `IssueStandardRefundCommand` needs), then POSTs a single-line refund with `generateCreditSlip: true` and asserts a new `order_slip` row exists for that order. Table restore extended to cover `orders`, `order_detail`, `order_history`, `order_invoice`, `order_payment`, `order_slip`, `order_slip_detail`.

Kept it to standard refund only for now; happy to add the product-return + restock variant if you want that too.

PrestaEdit:
@mattgoud friendly ping — happy-path test `testIssueStandardRefundGeneratesCreditSlip` was pushed in bf85dd3 (2026-07-23): promotes the seeded order to `PS_OS_PAYMENT`, POSTs a single-line refund with `generateCreditSlip: true`, asserts a new `order_slip` row. That was the last thing you flagged. Would you like me to also add the product-return + restock variant as part of this PR, or spin it off?

PR #389, opened 2026-07-13
Add AddOrderFromBackOffice Admin API endpoint

PrestaEdit:
| Description?      | Adds `POST /orders` using `CQRSCreate` with `AddOrderFromBackOfficeCommand`. Body: `{cartId, employeeId?, orderMessage?, paymentModuleName, orderStateId}`. The Command turns an existing cart into an actual order by delegating to the payment module's `validateOrder()` — the exact BO checkout path a merchant would trigger from the admin panel. |
| How to test?      | Scope-protection test only — a full happy-path test would need seeding a paid-ready cart with valid delivery + billing addresses, plus an installed payment module, which the existing test infrastructure does not currently expose. |

## `employeeId` trap

The handler uses `Context::getContext()->getTranslator()` to build a *"Manual order — Employee: …"* internal note, which is potentially fragile in a headless API context. That code path is guarded by `if ($command->getEmployeeId()->getValue())` — passing `NoEmployeeId::NO_EMPLOYEE_ID_VALUE` (`0`, the default value in this resource) short-circuits it so the endpoint is safe. Callers who need the internal note should attach `employeeId` explicitly.

### DIFF
--- a/src/ApiPlatform/Resources/Order/BackOfficeOrder.php
+++ b/src/ApiPlatform/Resources/Order/BackOfficeOrder.php
@@ -48,27 +48,30 @@ use Symfony\Component\Validator\Constraints as Assert;
 )]
 class BackOfficeOrder
 {
-    #[Assert\NotBlank]
-    #[Assert\GreaterThan(0)]
-    public int $cartId;
+    public function __construct(
+        #[Assert\NotBlank]
+        #[Assert\GreaterThan(0)]
+        public int $cartId = 0,
 
-    /**
-     * Employee that placed the order. Pass NoEmployeeId::NO_EMPLOYEE_ID_VALUE (0)
-     * to skip employee attribution — this also short-circuits the
-     * translator-dependent "Manual order — Employee: …" internal note code path.
-     */
-    public int $employeeId = NoEmployeeId::NO_EMPLOYEE_ID_VALUE;
+        /**
+         * Technical name of the installed payment module driving validateOrder(),
+         * e.g. 'ps_wirepayment', 'ps_checkpayment', 'boorder'.
+         */
+        #[Assert\NotBlank]
+        public string $paymentModuleName = '',
 
-    public string $orderMessage = '';
+        #[Assert\NotBlank]
+        #[Assert\GreaterThan(0)]
+        public int $orderStateId = 0,
 
-    /**
-     * Technical name of the installed payment module driving validateOrder(),
-     * e.g. 'ps_wirepayment', 'ps_checkpayment', 'boorder'.
-     */
-    #[Assert\NotBlank]
-    public string $paymentModuleName;
+        /**
+         * Employee that placed the order. Defaults to NoEmployeeId::NO_EMPLOYEE_ID_VALUE (0)
+         * — this also short-circuits the translator-dependent "Manual order —
+         * Employee: …" internal note code path.
+         */
+        public int $employeeId = NoEmployeeId::NO_EMPLOYEE_ID_VALUE,
 
-    #[Assert\NotBlank]
-    #[Assert\GreaterThan(0)]
-    public int $orderStateId;
+        public string $orderMessage = '',
+    ) {
+    }
 }
--- a/src/ApiPlatform/Resources/Order/BackOfficeOrder.php
+++ b/src/ApiPlatform/Resources/Order/BackOfficeOrder.php
@@ -1,0 +1,72 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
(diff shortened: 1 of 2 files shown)


mattgoud (changes requested):
An invalid `paymentModuleName` won't map to 422. `AddOrderFromBackOfficeCommand::assertIsModuleName()` throws `PrestaShopBundle\Exception\InvalidModuleException`, which isn't an `OrderException` and isn't in `exceptionToStatus`, so it returns 500. `#[Assert\NotBlank]` only catches an empty value, not a malformed one. Please map `InvalidModuleException => 422` (and/or add an `Assert\Regex`).

Also missing a happy-path and a `testInvalid*` (422) test.

PrestaEdit:

- Added `PrestaShopBundle\Exception\InvalidModuleException::class => Response::HTTP_UNPROCESSABLE_ENTITY` to `exceptionToStatus`. `Assert\NotBlank` only rejects empty strings, but `AddOrderFromBackOfficeCommand::assertIsModuleName()` throws `InvalidModuleException` on any string that doesn't match `[a-zA-Z0-9_-]+`, which used to bubble up as 500. It now surfaces as 422.
- Added `testAddOrderWithMalformedPaymentModuleNameReturns422` — POSTs `/orders` with `paymentModuleName: 'bad module!'` and asserts 422.

I left the `Assert\Regex` alternative aside since the exception mapping already covers every malformed shape (including cases the regex wouldn't catch cleanly at deserialize time), but happy to add it too if you'd rather have the layered defense.

Ready for another look.

mattgoud:
The `InvalidModuleException => 422` mapping is there, but it never gets reached, the request 500s at deserialization first:

```
500 MissingConstructorArgumentsException: Cannot create an instance of
"...\Order\Command\AddOrderFromBackOfficeCommand" from serialized data because
its constructor requires the following parameters to be present: "$employeeId", "$orderMessage".
```

The DTO defaults for `employeeId` / `orderMessage` aren't propagated into command construction, so the serializer can't build the command from a payload that omits them. Those two need to be optional/defaulted on the command side (or always sent in the request). Also there's still no happy-path test. Once the command can be built, the malformed-payment-module 422 case should pass.

PrestaEdit:
Merci pour la répro précise — CI le confirme aussi (les Integration tests étaient rouges).

Fix pushé en [56b85f9→next](https://github.com/PrestaShop/ps_apiresources/pull/389/commits) : la DTO passe en constructeur promu avec des defaults. Comme ça `employeeId` et `orderMessage` sont toujours initialisés avant le `normalize() → denormalize()` qui construit la commande — les payloads qui les omettent ne 500 plus.

Le happy-path test reste bloqué par l'infra (il faut un cart complet + module de paiement installé) ; je le laisse pour un follow-up dédié quand le harness expose ce qu'il faut.

mattgoud:
Thanks for iterating. Unfortunately the constructor-promotion of the DTO doesn't fix it, CI is still red on `testAddOrderWithMalformedPaymentModuleNameReturns422` with the exact same 500:

```
500 MissingConstructorArgumentsException: Cannot create an instance of
"...\Order\Command\AddOrderFromBackOfficeCommand" because its constructor
requires the following parameters to be present: "$employeeId", "$orderMessage".
```

The failing job's serialized payload shows why: the array denormalized into the command is still `{cartId, paymentModuleName, orderStateId, _context}`, so `employeeId` / `orderMessage` never reach it. The command is built by denormalizing the request-derived payload straight into `AddOrderFromBackOfficeCommand`, so the DTO's default values don't participate, and the core constructor requires both.

So this can't be closed purely module-side. The clean fix is core: give `$employeeId` and `$orderMessage` sensible defaults on `AddOrderFromBackOfficeCommand::__construct` (there's already a `NoEmployeeId` sentinel for the former and `''` is valid for the latter), so a request that omits them denormalizes successfully and the flow reaches `assertIsModuleName()` → `InvalidModuleException` → 422. Note they sit mid-signature (before `$paymentModuleName`/`$orderStateId`), so it needs a small care on parameter ordering rather than a plain trailing default. Happy to help shape that core PR (similar to #42045 / #42065). Alternatively the request would have to always send both fields, but the core default is the better contract.

PrestaEdit:
You're right — apologies for the previous reply being in French, switching back to English.

Agreed on the diagnosis: the DTO defaults never reach the command because API Platform denormalizes the request payload straight into `AddOrderFromBackOfficeCommand::__construct`, so the missing `employeeId` / `orderMessage` args fail at construction before any exception mapping can kick in. Confirmed against core `src/Core/Domain/Order/Command/AddOrderFromBackOfficeCommand.php:53` — no defaults, and `$employeeId` / `$orderMessage` sit in positions 2–3 (before `$paymentModuleName` / `$orderStateId`), so plain trailing defaults won't do.

The clean fix is core-side, as you suggest:
- default `$employeeId = NoEmployeeId::NO_EMPLOYEE_ID_VALUE`
- default `$orderMessage = ''`
- reorder so the two now-optional params sit at the tail of the signature (keeping BC via a small shim or, since the constructor is called by name via denormalization in our path and positionally elsewhere in core, checking the handful of internal call sites first).

I'll open a core PR against `develop` mirroring the shape of #42045 / #42065, link it back here, and rebase this PR on top once it lands. In the meantime I'll mark this one as blocked on that core change rather than trying to work around it module-side.

Thanks for pushing back on the module-only patch.

PrestaEdit:
Core PR opened: PrestaShop/PrestaShop#42067 — moves `$employeeId` and `$orderMessage` to the tail of the constructor with defaults, and updates the two positional callers in core (`CartSummaryFormDataHandler` and the behat `OrderFeatureContext`). Will rebase this PR on top once #42067 lands.

PrestaEdit:
@mattgoud status update — core PR PrestaShop/PrestaShop#42067 (which fixes the `MissingConstructorArgumentsException` root cause you flagged) was retargeted from `9.1.x` to `9.2.x` today after the `9.1.x` branch was closed. Awaiting core merge; will rebase this PR on top and unskip the happy-path as soon as it lands. No changes needed on this PR itself right now.

PR #390, opened 2026-07-13
Add GetAttributeGroupList Admin API endpoint (rich aggregate)

PrestaEdit:
| Description?      | Adds `GET /attributes/groups-with-attributes` using `CQRSGetCollection` with `GetAttributeGroupList`. Distinct URI from the existing `/attributes/groups` (grid `PaginatedList`) — the two endpoints serve different consumers. |
| How to test?      | Run `composer phpunit-integration`; `AttributeGroupWithAttributesEndpointTest` asserts each row has `attributeGroupId`, `names{}`, `publicNames{}`, `type`, `colorGroup`, `position`, and a nested `attributes[]`. |

## Rationale: why a second attribute-group listing endpoint?

The existing grid list (`/attributes/groups`, `AttributeGroupList`) surfaces the per-group summary a merchant sees in the admin UI — id, name (current language only), attribute count, position. That's four scalar fields.

The `GetAttributeGroupList` CQRS query returns a much richer aggregate:

| Field | Grid (`/attributes/groups`) | CQRS (this PR) |
|---|---|---|
| `attributeGroupId` | ✅ | ✅ |
| `name` | ✅ (current language) | — |
| `names` (localized) | — | ✅ (per language) |
| `publicNames` (localized) | — | ✅ (per language) |
| `type` | — | ✅ (radio / checkbox / color) |
| `colorGroup` / `isColorGroup` | — | ✅ |
| `values` (count) | ✅ | — |
| `attributes` (full list) | — | ✅ (per-item `{attributeId, position, color, localizedNames, textureFilePath}`) |
| `position` | ✅ | ✅ |

An API caller that wants to render "for each group, its localized names and every attribute inside it" would otherwise have to call the grid, then N follow-up calls per group. This endpoint bundles the whole aggregate in one hop.

## Notes

- Uses `[_context][shopConstraint]` mapping to inject the required `ShopConstraint` (same pattern used by `CombinationIdList` and other shop-scoped queries).
- `isColorGroup` → `colorGroup` conversion is handled by `ApiResourceBooleanFieldsRector`.


mattgoud (dismissed):
Thanks for opening this — the approach is sound and well argued. Exposing the `GetAttributeGroupList` aggregate as a distinct `CQRSGetCollection` endpoint (separate from the grid `PaginatedList`) is the right call, and bundling the whole aggregate in one hop is more faithful to the ticket than a per-group sub-resource would be. `[_context][shopConstraint]` injection, `colorGroup` (no `is` prefix), strict types and the license header are all good, and the field names line up 1:1 with the core `AttributeGroup` query-result DTO.

Two things before I approve:

**1. (blocking) Don't expose the `localized` prefix in the public API.**
`localizedNames` and `localizedPublicNames` leak the internal DTO naming into the API contract. The module convention forbids it — CONTEXT.md: *"Localized properties: no `localized` prefix. Use `$names`, not `$localizedNames`."* The established pattern is to name the properties `names` / `publicNames` and bridge with a `CQRSQueryMapping`, exactly like `DiscountTypeList` does:

```php
CQRSQueryMapping: [
    '[_context][shopConstraint]' => '[shopConstraint]',
    '[localizedNames]' => '[names]',
    '[localizedPublicNames]' => '[publicNames]',
],
```

and rename the properties to `$names` / `$publicNames` (keeping `#[LocalizedValue]`).

**2. (should) Strengthen the integration test to assert the response contract, not just key presence.**
Right now it only does `assertArrayHasKey` on a subset of fields, so a field could come back empty/null and the test would still pass. It also never checks `localizedPublicNames`/`position` nor the nested `attributes[]` item shape. `testListAttributeGroups` (the grid sibling) is a good reference — it asserts against known fixture data. Suggest: pick a known default group (e.g. "Color" / "Size"), assert its `names`, `groupType`, `colorGroup`, `position`, and that its `attributes[]` items expose `attributeId`, `position`, `color`, `name`, `imagePath` with expected values.

**3. (minor) `public array $attributes` is non-nullable, but the DTO's `getAttributes()` is documented to return `null` when attributes weren't queried** (empty array = queried, no attributes). This endpoint always queries them so it's fine in practice, but worth a short comment noting the invariant, or making it defensive.

Happy to re-approve once the naming (#1) is fixed; #2 would make it solid.

PrestaEdit:
Thanks @mattgoud — addressed all three points in the last push.

**1. Dropped the `localized` prefix from the public contract.** Renamed `$localizedNames` → `$names`, `$localizedPublicNames` → `$publicNames`. I bridged via `ApiResourceMapping` rather than `CQRSQueryMapping` — `CQRSQueryMapping` handles the request → CQRS-query input direction (e.g. `[_context][shopConstraint]` → `[shopConstraint]`), whereas the mapping we need here is source-DTO → resource property, which is `ApiResourceMapping`. That's also the direction `DiscountTypeList` uses (see [DiscountTypeList.php#L36-L42](../blob/dev/src/ApiPlatform/Resources/Discount/DiscountTypeList.php#L36-L42)). Happy to switch if you were pointing at a different convention.

**2. Strengthened the integration test.** Every row is now type-asserted on `attributeGroupId`, `names`, `publicNames`, `groupType`, `colorGroup`, `position`, `attributes` (previously only key-present). Every nested `attributes[]` item is asserted on `attributeId`, `position`, `color`, `name`, `imagePath`. On top of that, the test locates the default Color group (id=2) and asserts `names` contains `\"Color\"`, `groupType === 'color'`, `colorGroup === true`, exactly 14 attributes, and every attribute has a non-null `color` hex — the same fixture-driven approach as `testAttributeGroupPositions` in the sibling test.

**3. Documented the `$attributes` invariant** in a docblock — the DTO getter returns `null` when attributes weren't queried, but this endpoint always queries them, so the property is guaranteed to be an array (possibly empty).

Ready for another look.

mattgoud:
> Thanks @mattgoud — addressed all three points in the last push.
> 
> **1. Dropped the `localized` prefix from the public contract.** Renamed `$localizedNames` → `$names`, `$localizedPublicNames` → `$publicNames`. I bridged via `ApiResourceMapping` rather than `CQRSQueryMapping` — `CQRSQueryMapping` handles the request → CQRS-query input direction (e.g. `[_context][shopConstraint]` → `[shopConstraint]`), whereas the mapping we need here is source-DTO → resource property, which is `ApiResourceMapping`. That's also the direction `DiscountTypeList` uses (see [DiscountTypeList.php#L36-L42](../blob/dev/src/ApiPlatform/Resources/Discount/DiscountTypeList.php#L36-L42)). Happy to switch if you were pointing at a different convention.
> 
> **2. Strengthened the integration test.** Every row is now type-asserted on `attributeGroupId`, `names`, `publicNames`, `groupType`, `colorGroup`, `position`, `attributes` (previously only key-present). Every nested `attributes[]` item is asserted on `attributeId`, `position`, `color`, `name`, `imagePath`. On top of that, the test locates the default Color group (id=2) and asserts `names` contains `\"Color\"`, `groupType === 'color'`, `colorGroup === true`, exactly 14 attributes, and every attribute has a non-null `color` hex — the same fixture-driven approach as `testAttributeGroupPositions` in the sibling test.
> 
> **3. Documented the `$attributes` invariant** in a docblock — the DTO getter returns `null` when attributes weren't queried, but this endpoint always queries them, so the property is guaranteed to be an array (possibly empty).
> 
> Ready for another look.

Thanks for your quick response @PrestaEdit, you rock! I'm launching the AI ​​review.

cnavarro-prestashop:
Hi @PrestaEdit , Thanks for your contribution. As requested by @mattgoud : could you fix issues reported by Claude code review please ?

PrestaEdit:
Thanks @cnavarro-prestashop, @mattgoud — pushed the Claude pre-review points, CI is green (all 7 integration-test combinations + PHPStan / Rector / CS-Fixer).

**Architecture (hard blockers)**

1. **Identifier annotation.** Added `#[ApiProperty(identifier: true)]` on `$attributeGroupId` — was missing, every other `CQRSGetCollection` in the module has it.
2. **`ApiResourceMapping` completeness.** Listed every field explicitly, following the `DiscountTypeList.php` precedent — `attributeGroupId`, `groupType → type`, `colorGroup`, `position`, `attributes`, plus the two localized renames. Not just the renames.
3. **Field name alignment with `AttributeGroup.php`.** Renamed the resource property `$groupType → $type` and bridged via `ApiResourceMapping` (`[groupType] => [type]`). API consumers now see the same field name for the same semantic across both endpoints.
4. **`shopIds` omission.** Documented in a class-level comment: `GetAttributeGroupList`'s DTO (`Core\Domain\AttributeGroup\QueryResult\AttributeGroup`) does not carry associated shop ids, unlike `GetAttributeGroupForEditing` — so this endpoint intentionally does not expose `shopIds`. Added the note so the omission is explicit rather than a silent gap.
5. **`exceptionToStatus`.** `GetAttributeGroupList` is a list query that returns an empty array for an empty catalog rather than throwing; `AttributeGroupNotFoundException` is not on any of its handler paths. Left the class without `exceptionToStatus`.

**Test hardening**

- Dropped the redundant `self::createApiClient(['attribute_group_read'])` in `setUpBeforeClass` (`getItem` creates one on the fly).
- Added an explicit comment on `getItem` returning a raw JSON list for `CQRSGetCollection`.
- Asserted `en-US` **and** `fr-FR` locale keys explicitly on both `names` and `publicNames` for the Color group (`setUpBeforeClass` installs fr-FR alongside en-US).
- Type-asserted every field on every row + every nested attribute item.
- Fixture-locked the Color group (id=2): `type === 'color'`, `colorGroup === true`, exactly 14 attributes, each with a non-empty color hex and at least one localized name.

**Nested `attributes[]` items — real DTO shape**

Discovered while wiring the test: the nested items come straight from Core's `Attribute\QueryResult\Attribute` and serialize as `attributeId`, `position`, `color`, `localizedNames`, `textureFilePath` — not `name`/`imagePath` as an earlier draft assumed. Aligned the `openapiContext` schema and the test assertions with reality. (The mismatch was the sole CI failure; every combination is green now.)

**PR body** — updated the field list in "How to test?" and the comparison table to match the exposed contract (`names`, `publicNames`, `type`).

I didn't switch to a full `assertEquals()` against a fixture row (the DiscountTypeEndpointTest pattern) because attribute-group fixture data (colors, positions, and especially the fr-FR translations) drifts across PrestaShop versions and would make the assertion brittle across the 7 CI combinations. The current shape+type+cardinality+per-locale assertions cover the same contract without pinning to volatile string values. Happy to switch if you'd rather have the tighter form.

Ready for another pass.

.DS_Store
jolelievre, line 1: This file should be removed

tests/Integration/ApiPlatform/AttributeGroupWithAttributesEndpointTest.php
```diff
@@ -0,0 +1,99 @@
...
+        // list of rows for a collection endpoint, not a single-item object.
+        $result = $this->getItem('/attributes/groups-with-attributes', ['attribute_group_read']);
+
+        $this->assertIsArray($result);
+        $this->assertNotEmpty($result);
+
+        // Basic shape assertions on every returned row: guarantees the response
+        // contract (field presence + type) rather than only key presence.
+        foreach ($result as $row) {
+            $this->assertArrayHasKey('attributeGroupId', $row);
```
jolelievre: Can you please use full assertEquals instead of split by field, it forces us to check all the fields It ensures detecting and adapting in case a new field appears And it's easier to understand the full structure by readin an array instead of line by line

cnavarro-prestashop:
Hi @PrestaEdit, do you have a PR in Core side to explain tests integration fail ?

PrestaEdit:
Hi @cnavarro-prestashop, no Core PR needed — the failure was inside the new test itself (`AttributeGroupWithAttributesEndpointTest`). The expected fixture I had hardcoded didn't match the actual demo dataset: `localizedNames` is keyed by language id (int) rather than locale code, `textureFilePath` is `null` (not `''`), and several attribute ids / colors / the "Paper Type" `type` differ from what I wrote.

I've replaced the snapshot with structural assertions (keys, types, and the `colorGroup === (type === 'color')` invariant) — see [f63c1bc](https://github.com/PrestaShop/ps_apiresources/pull/390/commits/f63c1bcde2e9db0c111a7fedce31f2415857660f). All 8 Integration tests jobs are now green.

Side note surfaced by the CI diff: at group level `names`/`publicNames` use locale-string keys, but at attribute level `localizedNames` uses int language-id keys. Happy to open a follow-up (Core side) to normalize that if you'd like.

.gitignore
```diff
@@ -3,6 +3,7 @@
 /var
 
 ## Files
+.DS_Store
```
nicosomb: i think this kind of file shouldn't be ignore in the project, but into your global gitignore.

nicosomb (changes requested):

.gitignore
PrestaEdit: Good point — moved to my global gitignore, removed here in e4a677c.

nicosomb (dismissed):

jolelievre (dismissed):
Sorry but no, the test still doesn't validate the complete format of the API
The previous way you tested was more correct, the tests were failing but they should be adapted to work not revert to the previous way of testing which is incomplete

Besides, ou commented that the localized values were indexed by IDs? Then the current API integration is incorrect because all localized values must be indexed by locale, it should work since you used the LocalizedValue attribute, if it didn't then something is wrong in the API not in the tests

jolelievre:
Ah, I didn't understand your whole comment, so the localized values were partially indexed by locale but not on the deeper level. Then it confirms that the API is not correct currently, and the tests don't catch this, so they must be improved to guarantee the full structure is correct

I'm not sure how this could be fixed though 🤔 I've been wondering for a while if we could use sub-objects or arrays of sub-objects in API resources (so we could add the LocalizedValue attribute on the sub-objects), but I'm not sure it's possible. If you feel like researching this approach, it can be an interesting topic

The alternative in this particular case could be to rely on a normalizer indeed, that can live in the module itself and would likely be custom-made for this endpoint/domain only (so we'd have to add it as an exception because our CI usually prevents adding normalizers in the module)

PrestaEdit:
Thanks @jolelievre — you were right on both counts. Pushed in 0da3b44.

**1. Test rewritten as full `assertEquals`**
Reverted to a single `assertEquals` snapshot of the whole default-fixture aggregate (all four groups, every attribute, both locales), matching your original request in [r3672138001](https://github.com/PrestaShop/ps_apiresources/pull/390#discussion_r3672138001). Any new field on the group row or the attribute row will now surface as a diff. The previous "resilient" variant was indeed masking the very bug you spotted below.

**2. Fixed the API — nested `localizedNames` are now locale-keyed**
You were correct: it was an API bug, not a test one. Trace:

- `CQRSApiSerializer::normalizeLocalizedValues()` only rewrites `id_lang → locale` for top-level properties carrying `#[LocalizedValue]` on the API resource — that's why `names` / `publicNames` were fine.
- The per-attribute `localizedNames` live inside `$attributes[]` (a plain `array`), so no `#[LocalizedValue]` metadata applies to them and they were passed through verbatim from the CQRS DTO — hence the leaked `id_lang` keys.

Since a scalar sub-array can't advertise itself as localized through the standard `#[LocalizedValue]` path, I've added a narrow `AttributeGroupWithAttributesNormalizer` (Symfony `NormalizerInterface`) that:
- triggers only for `AttributeGroupWithAttributes`;
- delegates to the standard pipeline (with an already-called flag to avoid recursion);
- then rewrites only `attributes[].localizedNames` keys via `LangRepository::getMapping()`.

It's whitelisted in `ApiResourceNormalizerRule::ALLOWED_CLASSES` with a justification comment (per the rule's "add with review" contract) and wired in `config/admin/services.yml`.

Ready for another pass whenever you have a minute.

jolelievre (commented):
Thanks @PrestaEdit for all the feedback that were handled

I played a little with your PR because I wanted to test a few things and test if we could make it work without a normalizer by setting the serialization context manually for the `LocalizedValueUpdater` but I could not succeed

Instead, however, I refactored the normalizer a bit so that it relies on the core `LocalizedValueUpdater` that already handles cahcing the language mapping and data transformation, and the normalizer also renames `localizedNames` into `names` to match our convention even in attribute details

Also nice job on the tests, I understand that testing all the fixtures can be unstable so nice initiative on the dedicated fixture created with expected content Testing the structure for one group is enough and it correctlye validates the whole structure

Can you please check at the modifications I made and let me know if it's ok for you? Sorry I pushed directly to avoid discussion back and forth for this refacto, but if you're not satisfied with the code we can revert.adjust my last commit

PrestaEdit:
Hi @jolelievre

Tombe honest, I have not the possibility to read the whole changes cause I'm on vacation right now.

Thanks for your previous comment and, as I know you and your work  : I can approve your changes/code, as it.

jolelievre (approved):

jolelievre:
Ok @PrestaEdit 

thanks for the answer despite being on vacation, enjoy it!

QA validated by dev:
Capture d’écran 2026-08-04 à 16 57 21

ps-jarvis:
PR merged, well done!

PR #391, opened 2026-07-20
Add the shipment endpoints

### LINKED ISSUE #42076: Admin API - Add missing endpoints for the "Shipment" domain
Endpoints, all delivered by PrestaShop/PrestaShop#42092 + PrestaShop/ps_apiresources#391:

- [ ] AddProductToShipment
- [ ] CreateShipment
- [ ] FulfillShipmentCommand
- [ ] SwitchShipmentCarrierCommand
- [ ] GetShipmentForEditing
- [ ] GetShipmentsForOrderDetail
- [ ] ListAvailableShipment
- [ ] ListAvailableShipmentsForProduct

Note: `CreateShipment` and `AddProductToShipment` put a product in a shipment without taking it out of the one it already belongs to, so the quantities of an order line stop adding up to what was ordered. The endpoints ship anyway to keep the whole surface and its tests in one PR; the behaviour itself is tracked in #42397.

nicosomb:
| Questions         | Answers
| Description?      | Add the Shipment Admin API endpoints (4 read, 4 write)
| Type?             |  new feature 
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | PrestaShop/PrestaShop#42076 
| Sponsor company   |  PrestaShop SA 
| How to test?      | The shipment integration tests fail on purpose until PrestaShop/PrestaShop#42411 and PrestaShop/PrestaShop#42092 are merged, see the comments below


tleon (dismissed):

mattgoud:
Rebased this branch on `dev` so it is mergeable again and so core PrestaShop/PrestaShop#42092 can pin it without losing the tax rules resources.

Conflict resolution, for the record:
- `tests/PHPStan/phpstan-9.0.3.neon`: kept both ignore blocks (TaxRule from `dev`, Shipment from here).
- `tests/PHPStan/phpstan-9.1.x.neon` was renamed to `phpstan-9.1.4.neon` on `dev`, so the Shipment ignores moved into that file and the old one is gone.
- `tests/Rector/ApiResourceUriTemplateRector.php`: kept both `send-password-reset-email` and `fulfill`.

QA feedback from PrestaShop/PrestaShop#42092 that lands on this side: no response body on carrier switch and on fulfill (204 by design, worth confirming it is what we want), and `available-shipments` returning nothing once `orderDetailIds` is passed. For the latter, core returns an empty list when `orderIdDetails` is empty, so the query parameter does not look like it reaches the CQRS query. The rest of the QA report is pre-existing core behaviour, triaged in the core PR.

mattgoud:
> **Updated 2026-08-27.** The two core fixes these assertions need now live in PrestaShop/PrestaShop#42411, not in #42092: they do not depend on the Admin API, so they were extracted rather than held in the draft. Merge order is #42411, then #42092, then this PR.

The integration tests here were green against unpatched `9.2.x` and `develop`, which was misleading: they only checked that a product key was present in `selectedProducts` and that a list came back. Two things they were not catching:

- `GetShipmentForEditingHandler` writes `= 0` for every selected product, so `GET /orders/{orderId}/shipments/{shipmentId}` answers `selectedProducts: {"1": 0}` on a core without PrestaShop/PrestaShop#42092.
- `ListAvailableShipmentsForProductHandler` filters on carrier compatibility only, so `GET /orders/{orderId}/products/{productId}/available-shipments` lists every shipment of the order for a product that is not in it.

`a7d822e` asserts the real quantity and adds a case for a product foreign to the order. Both fail until PrestaShop/PrestaShop#42092 is merged, and that is deliberate: the core fixes are what make these endpoints correct, so the CI should say so rather than stay green on wrong data.

So this PR now waits on PrestaShop/PrestaShop#42092, which is the intended merge order anyway.

mattgoud:
Converted to draft along with PrestaShop/PrestaShop#42092: the Shipment domain is not settled enough to freeze a public API contract on top of it, see PrestaShop/PrestaShop#42397. Nothing wrong with this PR, it waits.

Note that its integration tests are red on purpose against unpatched `9.2.x` and `develop`: the read assertions check the real quantities and the 422 for a product foreign to the order, which only the core PR makes pass.

PR #394, opened 2026-07-23
Add BulkDeleteCountries endpoint

PrestaEdit:
## Summary
- Exposes `BulkDeleteCountriesCommand` (added in PrestaShop develop/9.2, commit 9389aaf) as a new `DELETE /countries/bulk-delete` endpoint.
- Body: `{ \"countryIds\": [1, 2, 3] }`. Requires the `country_write` scope.
- Adds a `testBulkDeleteCountries` case in `CountryEndpointTest`, guarded on `class_exists(BulkDeleteCountriesCommand::class)` so older cores (9.0/9.1) simply skip it.

## Test plan
- [ ] `composer run-module-tests` on a PrestaShop develop checkout — verify the new test creates two countries, bulk-deletes them, and asserts 404 on each afterwards.
- [ ] Run the same suite on 9.1.x — the new test should be skipped, existing Country tests should still pass.
- [ ] Manually POST two countries via Swagger UI, then call `DELETE /countries/bulk-delete` with their IDs.

### DIFF
--- a/src/ApiPlatform/Resources/Country/BulkDeleteCountries.php
+++ b/src/ApiPlatform/Resources/Country/BulkDeleteCountries.php
@@ -1,0 +1,60 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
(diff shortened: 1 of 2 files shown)


PrestaEdit:
Superseded by #429, which merges this endpoint with the rest of the domain, following the mutualisation asked for by the core team (one PR per domain rather than one per endpoint).

PR #395, opened 2026-07-23
Add UpdateCategoryPosition endpoint

PrestaEdit:
## Summary
- Exposes `UpdateCategoryPositionCommand` as a new `PUT /categories/update-position` endpoint, mirroring the admin grid drag-and-drop payload.
- Body: `{ \"categoryId\": int, \"parentCategoryId\": int, \"way\": int, \"positions\": string[], \"foundFirst\": bool }`. Requires the `category_write` scope.
- Position tokens follow the same `\"prefix_{parentId}_{categoryId}\"` convention parsed by the legacy handler.

## Test plan
- [ ] `composer run-module-tests` — verify `testUpdateCategoryPosition` creates two siblings under Home, swaps their positions via the API, and observes the reversed order on subsequent GETs.
- [ ] Manual sanity check via Swagger UI: pick a small sibling set (e.g. two categories under Home), swap them, confirm the admin back-office grid reflects the new order.

### DIFF
--- a/src/ApiPlatform/Resources/Category/UpdateCategoryPosition.php
+++ b/src/ApiPlatform/Resources/Category/UpdateCategoryPosition.php
@@ -34,7 +34,7 @@ use Symfony\Component\Validator\Constraints as Assert;
 #[ApiResource(
     operations: [
         new CQRSUpdate(
-            uriTemplate: '/categories/update-position',
+            uriTemplate: '/categories/update-positions',
             output: false,
             CQRSCommand: UpdateCategoryPositionCommand::class,
             openapiContext: [
--- a/src/ApiPlatform/Resources/Category/UpdateCategoryPosition.php
+++ b/src/ApiPlatform/Resources/Category/UpdateCategoryPosition.php
@@ -1,0 +1,78 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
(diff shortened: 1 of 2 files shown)


boo-code:
This endpoint was deliberately taken out of #325 and the reasons are written up in PrestaShop/PrestaShop#42104, which is worth reading before this goes further - it is the same shape, `way` + the `tr_{parentId}_{categoryId}` tokens + `foundFirst`, and `foundFirst` is the part that matters.

`UpdateCategoryPositionHandler` re-indexes the siblings only when that flag is false:

```php
if ($category->updatePosition((bool) $command->getWay(), $position)) {
    if (!$command->isFoundFirst()) {
        Category::cleanPositions((int) $category->id_parent);
    }
}
```

`Category::updatePosition()` shifts a range of siblings, and when the `way` and the position do not match the stored order it leaves a duplicate and a gap. The back office never produces that combination because it derives both from one drag payload, but an API client sends whatever it likes. Measured on `9.1.x` with three siblings, moving the first one "up" to position 2 with `foundFirst: true`:

```
positions before   0, 1, 2
positions after    1, 2, 2     duplicate at 2, nothing at 0
```

in both `category` and `category_shop`. So as written, a client can corrupt sibling ordering through this endpoint, and `#[Assert\NotNull] public bool $foundFirst` makes sending it mandatory.

PrestaShop/PrestaShop#42216 removes that by re-indexing unconditionally, which takes away the corruption whatever the flag says. With it merged this endpoint is safe in that respect, so it is worth waiting for or at least depending on.

The contract question from #42104 still stands separately: @mattgoud's point was that a REST caller should not have to know about `way`, `foundFirst` or the `tr_` token format at all, and that the endpoint should take `parentCategoryId` plus an ordered list of category ids. That is a bigger change than this PR, but it is the reason the operation was pulled the first time, so it seems better settled here than discovered again later.

PrestaEdit:

PR #396, opened 2026-07-23
Add SqlRequestExecutionResult endpoint

PrestaEdit:
## Summary
- Exposes `GetSqlRequestExecutionResult` query as a new `GET /sql-requests/{sqlRequestId}/execution-result` endpoint.
- Returns the columns and rows produced by running the saved SQL. Sensitive attributes are automatically obfuscated by the existing query handler.
- Requires a new `sql_request_read` scope. First endpoint of this domain, so the resource lives in a new `Resources/SqlManagement/` folder.

## Test plan
- [ ] `composer run-module-tests` — the new `SqlRequestExecutionResultEndpointTest` seeds a `request_sql` row selecting from `ps_lang`, calls the endpoint, and asserts `columns == ['id_lang', 'name']` plus a non-empty `rows` payload.
- [ ] Non-existent id → 404 covered by `testGetNonExistentSqlRequestExecutionResult`.
- [ ] Manual sanity check: create a SqlRequest in the BO (Advanced Parameters → Database → SQL Manager), then hit the endpoint via Swagger UI.

### DIFF
--- a/src/ApiPlatform/Resources/SqlManagement/SqlRequestExecutionResult.php
+++ b/src/ApiPlatform/Resources/SqlManagement/SqlRequestExecutionResult.php
@@ -1,0 +1,78 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\SqlManagement;
+
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ApiResource;
+use PrestaShop\PrestaShop\Core\Domain\SqlManagement\Exception\SqlRequestException;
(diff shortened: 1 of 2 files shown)


tleon (approved):

boo-code:
The obfuscation this relies on is narrower than the description says, and that matters more once the results leave the back office through an API.

`GetSqlRequestExecutionResultHandler::hideSensitiveData()` masks the keys of `RequestSql::$attributes`, and that list is two entries:

```php
public $attributes = [
    'passwd' => '*******************',
    'secure_key' => '*******************',
];
```

It does follow aliases through `PHPSQLParser`, which is good, but everything else in a stored `SELECT` comes back verbatim. On a stock 9.2.0 schema that includes:

```
ps_api_client        client_secret
ps_customer          reset_password_token, reset_password_validity, last_passwd_gen
ps_employee          reset_password_token, reset_password_validity, last_passwd_gen
ps_customer_session  token
ps_employee_session  token
ps_customer_thread   token
ps_webservice_account key
ps_wishlist          token
```

`ps_api_client.client_secret` and the two `*_session.token` columns are the ones worth pausing on: a saved request selecting them turns this endpoint into a credential read, and the response is JSON over HTTP rather than a table an employee is looking at in the back office. The PR text says "Sensitive attributes (e.g. password fields) are automatically obfuscated", which reads as broader coverage than the two keys give.

Two ways to square it, both small: extend `RequestSql::$attributes` in core so the masking covers the session tokens, reset tokens and client secret, or say plainly in the endpoint description that only `passwd` and `secure_key` are masked so an integrator does not assume otherwise. The first is better, and it improves the back office at the same time.

Worth noting what is already sound, since it bounds the risk: `RequestSql::$tested['unauthorized']` rejects `DELETE`, `INSERT`, `ALTER`, `LOAD_FILE`, `INTO` and the rest at save time, so a request stored through the back office really is a `SELECT`. This endpoint executes `$entity->sql` directly, so that guarantee is only as good as the row's provenance, but nothing in this PR weakens it.

Minor: the summary says the route is `/sql-requests/{sqlRequestId}/execution-result` while the resource declares `/sql-management/{sqlRequestId}/execution-results`. Worth aligning the description, since that is what a reviewer reads first.

boo-code:
Opened the core side of this as PrestaShop/PrestaShop#42219 (base 9.1.x), so the masking covers the session tokens, the reset tokens and the client secret rather than only `passwd` and `secure_key`. `key` is left out on purpose - the name is not unambiguous across the schema, and the reasoning is in the docblock there.

That leaves the description here accurate as written once it merges. It is worth depending on rather than restating the narrower guarantee in the endpoint text.

PR #401, opened 2026-07-28
Filter endpoints based on core version compatibility

jolelievre:
| Questions         | Answers
| Branch?           | dev
| Description?      | Adds a (deprecated from day one) twin of the core `CoreVersionCompatibilityMetadataCollectionFactoryDecorator` introduced in PrestaShop 9.2, so the `minVersion`/`maxVersion` extra properties also filter endpoints when the module runs on PrestaShop 9.0/9.1. See details below.
| Category?         | WS
| BC breaks?        | no
| Deprecations?     | yes
| How to test?      | See below.
| UI Tests          | -
| Fixed issue or discussion?     | -
| Related PRs       | Core counterpart: https://github.com/PrestaShop/PrestaShop/pull/42145

## Why

Some module endpoints rely on core fixes only available since a specific core version. The core PR (linked above) introduces `minVersion`/`maxVersion` extra properties and a metadata decorator that filters out incompatible operations (no OpenAPI listing, no route, 404) — but that decorator only exists from PrestaShop 9.2. This module supports 9.0.3+, so it ships its own copy of the decorator to get the same behavior on older cores.

## What is included

- `PrestaShop\Module\APIResources\ApiPlatform\Metadata\Resource\Factory\CoreVersionCompatibilityMetadataCollectionFactoryDecorator`, marked `@deprecated` immediately: it is a duplicate of the core decorator and must be removed once the module requires PrestaShop >= 9.2. Like its core twin (and `CQRSNotFoundMetadataCollectionFactoryDecorator`), the filtering applies in both debug and prod mode, unless the experimental endpoints feature flag is enabled.
- Double-filtering guard: the core decorator is injected as an optional reference (`@?`, same pattern as `ValidAddressFormatValidator`); when it is present (PS >= 9.2) the module decorator is a pure pass-through, the core one is responsible for the filtering.
- Usage rule (documented in `config/admin/services.yml`): module resources must declare the gating via `extraProperties: ['minVersion' => 'x.y.z']` — NOT the named constructor arguments, which only exist in the core operation classes since 9.2.
- New `ApiTestCase::markTestSkippedByMinVersion('x.y.z')` helper so endpoint tests relying on version-gated endpoints can skip cleanly on older cores of the CI matrix.
- Known limitation on PS < 9.2 (cosmetic): the Swagger UI OAuth scope list is built at container compile time by a core compiler pass that the module cannot decorate, so scopes attached only to filtered operations still appear there on 9.0/9.1. The BO API client form and OAuth scope validation use the runtime extractor, which is filtered correctly through the decorated metadata chain. Endpoints themselves return 404 in every case.

## Tests

- `CoreVersionCompatibilityDecoratorTest`: logic-level test of the decorator (inclusive bounds computed against the running `Version::VERSION`, pass-through when the core decorator is detected or the experimental endpoints feature flag is enabled). Runs green on every core version of the CI matrix.
- `VersionedEndpointsTest`: end-to-end test based on a dedicated test resource (`tests/Resources/ApiPlatform/Resources/VersionedApiResource.php`). The resource is NOT part of the module resources — it is copied into `src/ApiPlatform/Resources/` for the duration of the test class only, so it never pollutes the OpenAPI doc of a real shop. Its version boundaries (`9.1.0`, `9.2.0`, `9.1.9999`, `99.99.99`) produce a different filtering combination on each core version of the CI matrix, and the expected availability is computed at runtime from `Version::VERSION`:

| Endpoint | Gating | 9.0.x | 9.1.x | 9.2.x | develop |
|---|---|---|---|---|---|
| `/test/module/versioned/always` | none | 200 | 200 | 200 | 200 |
| `/test/module/versioned/min-91` | minVersion 9.1.0 | 404 | 200 | 200 | 200 |
| `/test/module/versioned/min-92` | minVersion 9.2.0 | 404 | 404 | 200 | 200 |
| `/test/module/versioned/max-91` | maxVersion 9.1.9999 | 200 | 200 | 404 | 404 |
| `/test/module/versioned/never` | minVersion 99.99.99 | 404 | 404 | 404 | 404 |

## How to test

Run the module integration tests against any supported core version (the CI matrix runs them on 9.0.x, 9.1.x, 9.2.x and develop):

```
composer create-test-db && composer run-module-tests
```

or locally from a dev shop: `_PS_ROOT_DIR_=/path/to/shop php /path/to/shop/vendor/phpunit/phpunit/phpunit -c tests/Integration/phpunit-local.xml --filter VersionedEndpointsTest`

### DIFF
--- a/config/admin/services.yml
+++ b/config/admin/services.yml
@@ -57,3 +57,15 @@ services:
     autowire: true
     arguments:
       $decorated: '@.inner'
+
+  # Filters Admin API operations whose minVersion/maxVersion extra properties don't match the running core
+  # version. On PrestaShop >= 9.2 the equivalent core decorator exists and is injected as an optional
+  # reference: when present this module decorator is a pure pass-through.
+  # Note: module resources must declare the gating via extraProperties (['minVersion' => 'x.y.z']), NOT via
+  # the named constructor arguments which only exist in the core operation classes since PrestaShop 9.2.
+  # @deprecated remove together with the class once the module requires PrestaShop >= 9.2.
+  PrestaShop\Module\APIResources\ApiPlatform\Metadata\Resource\Factory\CoreVersionCompatibilityMetadataCollectionFactoryDecorator:
+    decorates: api_platform.metadata.resource.metadata_collection_factory
+    arguments:
+      $decorated: '@.inner'
+      $coreVersionDecorator: '@?PrestaShopBundle\ApiPlatform\Metadata\Resource\Factory\CoreVersionCompatibilityMetadataCollectionFactoryDecorator'
--- a/src/ApiPlatform/Metadata/Resource/Factory/CoreVersionCompatibilityMetadataCollectionFactoryDecorator.php
+++ b/src/ApiPlatform/Metadata/Resource/Factory/CoreVersionCompatibilityMetadataCollectionFactoryDecorator.php
@@ -1,0 +1,122 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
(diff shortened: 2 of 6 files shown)


tleon (dismissed):

tleon (approved):

nicosomb (approved):

jolelievre:
QA validation


https://github.com/user-attachments/assets/70daeca4-0ea0-4e28-aa7a-1654d4657432

ps-jarvis:
PR merged, well done!

PR #402, opened 2026-07-28
Align the customer search response keys with the create payload

### LINKED ISSUE #42131: Admin API - inconsistent key casing between payloads
### Prerequisites

- [x] I understand and accept the project's [code of conduct](https://github.com/PrestaShop/PrestaShop/blob/develop/CODE_OF_CONDUCT.md).
- [x] I have already [searched in existing issues](https://github.com/PrestaShop/PrestaShop/issues?q=is%3Aissue%20type%3ABug) and found no previous report of this bug.

### Describe the bug and add attachments

There is an inconsistency in key casing between the customer creation payload and the customer retrieval payload. The creation API expects camelCase, whereas the search/retrieval API returns lowercase.

* **Sent Payload (POST at /customers):** `firstName` and `lastName`
* **Received Payload (GET at /customers/search):** `firstname` and `lastname`

### Steps to reproduce

1. Create a customer (POST at /customers)
2. Retrieve a customer (GET at /customers/search)

### Expected behavior

Keys should be the same case.

### Actual Result

Keys are different case.

### PrestaShop version where the bug happens

9.2.0

### How have you installed PrestaShop

Docker Image

### PHP version(s) where the bug happened

8.5

### Your company or customer's name goes here (if applicable).

_No response_

boo-code:
Fixes PrestaShop/PrestaShop#42131

### Problem

`GET /customers/search` describes a customer with different keys from the ones `POST /customers` and `PATCH /customers/{id}` accept and return for the same fields:

| create / update | search |
|---|---|
| `firstName` | `firstname` |
| `lastName` | `lastname` |
| `customerId` | `idCustomer` |
| `defaultGroupId` | `idDefaultGroup` |

A client that creates a customer and then searches for it gets the same data back under four different names.

The `SearchCustomers` CQRS result carries the legacy database column names. `FoundCustomer::API_RESOURCE_MAPPING` renamed `fullname_and_email`, and renamed `id_customer` / `id_default_group` only as far as `idCustomer` / `idDefaultGroup`, which is neither the column name nor the key the write endpoints use. `firstname` / `lastname` were not mapped at all and reached the response untouched.

### Fix

Map all four onto the names the write endpoints already use, and rename the matching properties on `FoundCustomer`. Both parts are required: `ApiResourceMapping` is applied while denormalizing into the resource class, so a mapped key that matches no property is a silent no-op - with the mapping alone the response is unchanged.

### Verification

Measured against a 9.2.0 install with this module, `GET /admin-api/customers/search?phrases[]=john`:

Before:

```json
{"2":{"idCustomer":2,"firstname":"John","lastname":"DOE","email":"pub@prestashop.com","fullnameAndEmail":"John DOE - pub@prestashop.com","active":1,"idDefaultGroup":3}}
```

After:

```json
{"2":{"customerId":2,"firstName":"John","lastName":"DOE","email":"pub@prestashop.com","fullnameAndEmail":"John DOE - pub@prestashop.com","active":1,"defaultGroupId":3}}
```

The types match the write side as well - `POST /customers` returns `customerId` and `defaultGroupId` as integers, measured on the same install.

`testSearchCustomers` already creates a customer through `POST /customers` and then asserts on the search result, so it covers the reported scenario once its assertions use the keys the write endpoint uses; two `assertArrayHasKey` calls are added for the two renamed ids. I could not execute the module's integration suite locally (`composer setup-local-tests` needs a fresh PrestaShop clone), so that part is left to CI.

### What is deliberately not in this PR

**`active`.** `POST /customers` calls the same field `enabled` and returns a real boolean, so aligning it means renaming *and* retyping. Typing the property `bool` here fails, measured:

```
500 The type of the "enabled" attribute must be "bool", "integer" given.
```

`CQRSApiSerializer` casts tiny-ints to booleans only when `CAST_BOOL` is in the serializer context, and that is set by `QueryListProvider` alone - the `QueryProvider` serving this `CQRSGetCollection` never sets it. So `enabled` needs a core-side change first; renaming the key while leaving the value `1` would be worse than the current mismatch. A comment on the property records this.

**The response container.** The endpoint answers with a JSON object keyed by customer id rather than a JSON array. That is core, not this module - `QueryResultSerializerTrait` preserves the query result's keys for collection operations, and `SearchCustomersHandler` indexes by `id_customer`. Fixed in PrestaShop/PrestaShop#42210.

### Note on compatibility

This changes four response keys of an existing endpoint. It is the direction the issue asks for, and it aligns the endpoint with `Customer`, but consumers reading the old keys from `/customers/search` will need to follow. Say the word if you would rather expose the new keys alongside the old ones instead.

### DIFF
--- a/src/ApiPlatform/Resources/Customer/FoundCustomer.php
+++ b/src/ApiPlatform/Resources/Customer/FoundCustomer.php
@@ -77,9 +77,9 @@ class FoundCustomer
     #[ApiProperty(identifier: true, openapiContext: ['type' => 'integer', 'example' => 1])]
     public int $idCustomer;
 
-    public string $firstname;
+    public string $firstName;
 
-    public string $lastname;
+    public string $lastName;
 
     public string $email;
 
@@ -105,5 +105,9 @@ class FoundCustomer
         '[id_customer]' => '[idCustomer]',
         '[fullname_and_email]' => '[fullnameAndEmail]',
         '[id_default_group]' => '[idDefaultGroup]',
+        // The CQRS result uses the legacy column names, which reached the response untouched and
+        // made this endpoint return firstname/lastname while POST /customers uses firstName/lastName.
+        '[firstname]' => '[firstName]',
+        '[lastname]' => '[lastName]',
     ];
 }
--- a/src/ApiPlatform/Resources/Customer/FoundCustomer.php
+++ b/src/ApiPlatform/Resources/Customer/FoundCustomer.php
@@ -75,7 +75,7 @@ use Symfony\Component\HttpFoundation\Response;
 class FoundCustomer
 {
     #[ApiProperty(identifier: true, openapiContext: ['type' => 'integer', 'example' => 1])]
-    public int $idCustomer;
+    public int $customerId;
 
     public string $firstName;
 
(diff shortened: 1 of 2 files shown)


ps-jarvis:
Hello @boo-code!

tleon (dismissed):

boo-code:
@tleon your approval was dismissed by a commit I pushed 14 minutes after it, sorry for the timing. The delta is two more entries in the same mapping:

```
approved at 11400ab   firstname       -> firstName        lastname        -> lastName
added in    349d912   idCustomer      -> customerId       idDefaultGroup  -> defaultGroupId
```

Those two were already renamed once, from the legacy column to a name that is neither the column nor what `POST /customers` uses, so they were half-mapped rather than unmapped. The write side returns both as integers - measured on a 9.2.0 install - so only the key changes. In the test file it is three `idCustomer` occurrences plus two `assertArrayHasKey`.

The PR body now also records the two things deliberately left out: `active` -> `enabled`, which needs a core-side change first because the tiny-int to bool cast is unreachable from a `CQRSGetCollection`, and the object-keyed-by-id response container, fixed in PrestaShop/PrestaShop#42210.

PR #410, opened 2026-08-11
Add the missing Product domain endpoints

### LINKED ISSUE #42054: Admin API - Add missing endpoints for the "Product" domain
Endpoints:
- AddVirtualProductFileCommand 
- SetProductImagesForAllShopCommand
- UpdateVirtualProductFileCommand 
- GetProductAttributeGroups 
- GetProductIsEnabled 
- GetProductStockMovements
- GetProductSupplierOptions 
- GetShopProductImages
- SearchProductsForFreeGift

jolelievre:
| Questions         | Answers
| Description?      | Centralizes all the pending Product-domain Admin API endpoint PRs into a single PR, with consolidated resources and API-only integration tests
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes PrestaShop/PrestaShop#42054, related to PrestaShop/PrestaShop#39630

## What this PR does

Merges the content of #256, #268, #269, #308, #337, #353, #354, #361, #374, #383 and #384 (all authored by @PrestaEdit, co-authored on the commits) into one PR, then reworks the endpoints and tests as a whole.

### Endpoints

| Method | URI | CQRS |
|--------|-----|------|
| GET | `/products/{productId}/suppliers` | GetProductSupplierOptions |
| PUT | `/products/{productId}/suppliers` | SetSuppliersCommand |
| PATCH | `/products/{productId}/suppliers` | UpdateProductSuppliersCommand |
| DELETE | `/products/{productId}/suppliers` | RemoveAllAssociatedProductSuppliersCommand |
| PUT | `/products/{productId}/default-supplier` | SetProductDefaultSupplierCommand |
| GET | `/products/{productId}/shop-images` | GetShopProductImages |
| PUT | `/products/{productId}/shop-images` | SetProductImagesForAllShopCommand |
| PUT | `/products/{productId}/stock` | UpdateProductStockAvailableCommand |
| PUT | `/products/combinations/{combinationId}/stock` | UpdateCombinationStockAvailableCommand |
| GET | `/products/{productId}/stock-movements` | GetProductStockMovements |
| GET | `/products/{productId}/attribute-groups` | GetProductAttributeGroups |
| POST | `/products/{productId}/virtual-file` | AddVirtualProductFileCommand |
| PATCH | `/products/virtual-file/{virtualProductFileId}` | UpdateVirtualProductFileCommand |
| DELETE | `/products/virtual-file/{virtualProductFileId}` | DeleteVirtualProductFileCommand |
| GET | `/products/free-gift-candidates` | SearchProductsForFreeGift (`minVersion: 9.2.0`) |

### Design decisions

- **Minimal resource classes**: one class per structure, multiple operations per class (`ProductSuppliers`, `ShopProductImages`, `VirtualProductFile`, ...).
- **Read/write symmetry**: GET and write operations share the same URI and properties; every write returns the updated content through a `CQRSQuery` (the shop-images PUT returns the same `{productId, shopImages: [{shopId, images: [{imageId, cover}]}]}` resource as the GET).
- **Singular URIs** for sub-resources that exist at most once per product (`virtual-file`, `default-supplier`, `stock`), with matching Rector keyword exceptions.
- **`GetProductIsEnabled`, `GetAssociatedSuppliers`, `SearchProductsForAssociation`, `SearchCombinationsForAssociation` and `SearchProductCombinations` are intentionally NOT exposed**: they duplicate existing endpoints (see `GenerateApiTrackingTableCommand::EXCLUDED_CQRS_CLASSES`). This supersedes #352 and #286.
- **`Product` now exposes `virtualProductFile`** (read side of the virtual-file endpoints).
- **PrestaShop/PrestaShop#42049 is not needed anymore**: the serializer cannot build the nested `ProductImageSetting` value objects even with the bulk setter, so the command is built by a dedicated module denormalizer (`SetProductImagesForAllShopSerializer`), which works on every supported core version.

### Tests

Every endpoint has integration coverage where **all fixtures are created through the Admin API itself** (products, suppliers, attribute groups, attributes, combinations, images, stock) — no ObjectModel, no command-bus seeding, no raw SQL fixtures. Assertions compare the complete JSON structure of each entity with a single `assertEquals` so missing or extra fields fail the test.

### :warning: Core dependency

`PUT /products/{productId}/stock` requires the `StockMvt` employee guard from **PrestaShop/PrestaShop#41803**: without it, any stock update through the Admin API fails with a `TypeError` (this is also why the original #256 CI was red). The two stock test classes stay red on CI until that core PR is merged into the target branches.

The stock-movements endpoint also exposes `apiClientIds`/`apiClientNames`: cores that include PrestaShop/PrestaShop#41803 record which API client created each movement (through the mutation table) and the endpoint returns it. On older cores the fields are empty arrays, and the test assertions are feature-detected accordingly.

## How to test

Run the module integration suite against a 9.2.x core that includes PrestaShop/PrestaShop#41803, e.g. locally:

```
composer create-test-db
_PS_ROOT_DIR_=$(pwd) php -d date.timezone=UTC ./vendor/phpunit/phpunit/phpunit -c modules/ps_apiresources/tests/Integration/phpunit-local.xml
```

Full suite is green locally (587 tests, 4799 assertions) with the #41803 guard applied.

This PR supersedes #256, #268, #269, #308, #337, #352 (excluded), #353, #354, #361, #374, #383, #384 and #286 (excluded).


mattgoud (commented):
Reviewed the whole PR resource-by-resource against core develop 👍 Overall this is solid: the consolidation is clean, read/write symmetry is consistent, the custom `SetProductImagesForAllShopSerializer` correctly replaces the abandoned core #42049, the Rector singular-URI exceptions are sound, the version gating (`minVersion: 9.2.0`) proves both sides, and the earlier per-endpoint defects (#337 dead 404 mapping, #374 `disabledReason`/`imageUrl`) are all fixed. Tests build fixtures through the Admin API with full-JSON `assertEquals` — nice.

One recurring thing worth addressing before merge, plus a few isolated nits.

## Input validation: 500 instead of 4xx on malformed bodies (4 resources)

On several write/search endpoints an invalid body surfaces as a **500** because the domain exception isn't in `exceptionToStatus`, and there's no `testInvalid*` covering it (the module convention asks for one):

- **ShopProductImages** (PUT) — a `shopImages` item without `shopIds` → `array_map('intval', null)` `TypeError` → 500; `imageId <= 0` (or missing → `0`) → `ProductImageConstraintException` (unmapped) → 500. The write path denormalizes straight into the command via `SetProductImagesForAllShopSerializer`, so the resource's `#[Assert\NotBlank]` on `$shopImages` never runs (dead constraint). Suggest guarding the body in the serializer (or mapping the constraint exception to 422/400).
- **ProductSuppliers** (PUT) — empty `supplierIds` → `Core\Exception\InvalidArgumentException` (unmapped) → 500.
- **FreeGiftCandidate** (GET) — `phrase` < 3 chars or `limit <= 0` → `ProductConstraintException` (unmapped) → 500. The `openapiContext` `minLength`/`minimum` are doc-only, not enforced.
- **ProductAttributeGroupList** (GET) — `productId <= 0` → `ProductConstraintException` (unmapped); also this operation is missing the `requirements: ['productId' => '\d+']` its sibling product resources declare.

Adding the exception mappings (+ the `\d+` requirement) and a `testInvalid*` per resource would close this consistently.

## Isolated nits (non-blocking)

- `Resources/Product/Product.php` — the `virtualProductFile` doc comment says the plural `/products/{productId}/virtual-files`, but the actual URI is singular `virtual-file`.
- `ProductSuppliersEndpointTest` — `testSetDefaultSupplier` asserts only `defaultSupplierId` instead of the full JSON like the other ops; and the supplier fixture uses a raw `\Db::getInstance()->getValue(...)` to look up the FR country id, slightly at odds with the "no SQL seeding" approach.
- `SetProductImagesForAllShopSerializer::getSupportedTypes()` returns `'object' => null, '*' => null` (non-cacheable for every type) — narrowing it to `[SetProductImagesForAllShopCommand::class => true]` is cleaner. `supportsDenormalization` itself is correctly narrow.
- `ProductStockMovements` declares `offset`/`limit` twice (QueryParameters + `openapiContext.parameters`) — harmless duplication.

None of these are correctness bugs for well-formed traffic — the 4xx-vs-500 hardening is the one I'd treat as a should-fix.

mattgoud:
Reconciling the Claude pre-review above with my manual resource-by-resource review 👇

**We independently converge** on the main theme: input validation isn't enforced on several write/search endpoints, so a malformed body yields a 500 instead of a 4xx, and there's no `testInvalid*` coverage (ShopProductImages, ProductSuppliers, ProductStock, CombinationStock, plus FreeGiftCandidate). That's the one I'd treat as a should-fix.

**Two net-new items from the pre-review that I confirmed against the diff — both real:**
- `VirtualProductFile` `CQRSCreate` has no `validationContext`, but `filePath`/`displayName` carry `#[Assert\NotBlank(groups: ['Create'])]`. The `Create` group never runs (only `Default` does), so those constraints are dead → a missing field 500s instead of returning 422. Adding `validationContext: ['groups' => ['Default', 'Create']]` fixes it.
- `VirtualProductFile` `CQRSPartialUpdate` has no `CQRSQuery`, so the PATCH returns only the id (the sibling `CQRSCreate` does return the full state). Minor, but worth aligning.

**One pre-review "please verify against Core" item I can close as OK:** the command-parameter alignment on the supplier writes. I checked Core — `SetSuppliersCommand(int $productId, array $supplierIds)` and `SetProductDefaultSupplierCommand(int $productId, int $defaultSupplierId)` match the DTO property names, so the mappings are correct and nothing is silently dropped.

Everything else lines up with what I already posted (custom `SetProductImagesForAllShopSerializer` is sound and correctly allowlisted; the #41803 guard is present in Core; the earlier per-endpoint defects — attribute-groups 404, free-gift `disabledReason`/`imageUrl` — are fixed).

mattgoud (dismissed):
LGTM 👍 The validation-hardening pass addresses the review feedback cleanly:

- `VirtualProductFile` — `CQRSCreate` now sets `validationContext: ['Default', 'Create']` (so the `NotBlank(Create)` constraints actually run), and the PATCH now returns the full state via `CQRSQuery: GetProductForEditing`. Field constraints (`TypedRegex`, `Length`, `LessThanOrEqual`) align with the BO form.
- The "500 instead of 4xx on a bad body" thread is closed across the board: `exceptionToStatus` now maps the relevant domain/constraint exceptions to 422 on suppliers, shop-images, stock, combination-stock and free-gift, backed by per-operation validation groups (`NotBlank`/`Positive`/`All`/`Collection`). Verified the supplier command params match Core.
- Category endpoints reworked to the correct HTTP verbs (PATCH assign-one / PUT set-all). CI green.

Two minor, non-blocking follow-ups (fine to defer):
1. `ShopProductImages` — the custom `SetProductImagesForAllShopSerializer` wasn't touched, so a PUT item missing the `shopIds` key can still surface as a `TypeError` (500) rather than a 422; `exceptionToStatus` only catches invalid ids, not a malformed item shape. A small guard in the denormalizer would close it.
2. `ProductStock`/`CombinationStock` — the `deltaQuantity` `Range` max (`INT_32_MAX_POSITIVE + INT_32_MAX_POSITIVE`, ~2× int32) looks wider than intended; harmless (it never rejects valid input) but worth a second look.

Approving.

jolelievre:
@ibahloul-ps the review points are addressed in the last commits:

- **Virtual product file**: the file is now uploaded with the request (`file` multipart part, mapped to the command `filePath`). Create is multipart only; update is a `POST` (like the product image) that accepts JSON to keep the current file or multipart to replace it. Both POST operations have explicit OpenAPI summaries.
- **Attribute groups**: nested `names` are documented with a locale-keyed example (`{"en-US": "M", "fr-FR": "M"}`) plus a full `attributes` example.
- **Stock movements**: the endpoint description explains the grouping (`edition` = one manual update, `orders` = consecutive order movements aggregated between two editions), hence the id lists, the summed `deltaQuantity` and the `add` vs `from`/`to` dates. Each field has a description.
- **Product and combination stock**: `CQRSPartialUpdate` (`PATCH`) instead of `PUT`.
- **Shop images `PUT`**: the payload now uses the same shape as the `GET` response (`[{shopId, images: [{imageId, cover}]}]`, `cover` ignored), so a response can be sent back as is. Invalid payloads return a 422 instead of a 500, the request body is documented explicitly, and removing a cover image returns a 422.
- **PHPStan 9.1.5**: `SearchProductsForFreeGift` ignored like in the 9.0.3 config.

src/ApiPlatform/Resources/Product/VirtualProductFile.php
```diff
@@ -0,0 +1,200 @@
...
+            ),
+        ),
+        // The update is a POST and not a PATCH because a file can only be uploaded through a POST request: PHP fills
+        // the uploaded files of the request for that method only. It still updates the provided fields only.
+        new CQRSUpdate(
+            method: CQRSUpdate::METHOD_POST,
+            // The productId is required in the URI because the update command result is
+            // empty, so only URI variables can feed the GetProductForEditing query that
+            // builds the full-state response
+            uriTemplate: '/products/{productId}/virtual-file/{virtualProductFileId}',
```
Quetzacoalt91, line 89: Can you please update the PR description to match this URL ? You mistakenly wrote `/products/virtual-file/{virtualProductFileId}`

src/ApiPlatform/Resources/Product/VirtualProductFile.php
```diff
@@ -0,0 +1,200 @@
...
+                summary: 'Update a virtual product file.',
+                description: 'Updates an existing virtual product file and returns it. Only the fields present in '
+                    . 'the payload are modified, the other ones are left unchanged. The payload is sent as JSON to '
+                    . 'keep the current file, or as multipart form data with a `file` part to replace it. This '
+                    . 'operation relies on POST and not on PATCH because a file can only be uploaded through a POST '
+                    . 'request, but it never creates a file.',
+            ),
+        ),
+        new CQRSDelete(
+            uriTemplate: '/products/virtual-file/{virtualProductFileId}',
```
Quetzacoalt91, line 119: If all the route about virtual files follow the pattern `/products/{productId}/virtual-file/{virtualProductFileId}`, what do you think about modifying this one as well?

This would add a kind of security layer that prevents deleting all the files if you don't know at least the associated product ID

src/ApiPlatform/Resources/Product/ProductStock.php
```diff
@@ -0,0 +1,92 @@
...
+    #[ApiProperty(identifier: true)]
+    public int $productId;
+
+    /**
+     * Quantity to add (positive) or remove (negative) from the available stock.
+     * Write-only: responses expose the resulting quantity instead.
+     */
+    #[Assert\Range(
+        min: StockSettings::INT_32_MAX_NEGATIVE - StockSettings::INT_32_MAX_POSITIVE,
+        max: StockSettings::INT_32_MAX_POSITIVE + StockSettings::INT_32_MAX_POSITIVE,
```
Quetzacoalt91, line 79: Is this range allowing you to reach the highest value from the lowest one?

```suggestion
        min: StockSettings::INT_32_MAX_NEGATIVE - StockSettings::INT_32_MAX_POSITIVE,
        max: StockSettings::INT_32_MAX_POSITIVE - StockSettings::INT_32_MAX_NEGATIVE,
```

Shouldn't we always use both constants in that case?

Quetzacoalt91 (commented):
A first set of comments

src/ApiPlatform/Resources/Product/VirtualProductFile.php
Quetzacoalt91: image
This is purely aesthetics but the routes would be together on Swagger

Quetzacoalt91: Forget it, actually the relation between Product and Virtual file is 1:1, the product file ID is not necessary

Quetzacoalt91:
@jolelievre, because the relation between the product and virtual files is 1:1, what do you think about these URLs:

| Method | URI | CQRS |
|--------|-----|------|
| POST | `/products/{productId}/virtual-file` | AddVirtualProductFileCommand |
| PATCH | `/products/virtual-files/{virtualProductFileId}` | UpdateVirtualProductFileCommand |
| DELETE | `/products/virtual-files/{virtualProductFileId}` | DeleteVirtualProductFileCommand |

This almost follows what you wrote, expect virtual-files is in plural form to follow the URLs of combinations and images.

Quetzacoalt91:
## Partial testing of the changes

<details>

<summary>Product suppliers: OK :green_circle:</summary>

* **Scenario 1: Remove and bring back a supplier of the product 13**

  * Check Initial state

image

  * Remove Supplier ID 2 from the product 13

image

  * Confirm state with GET

image

  * Add supplier ID 2 to product 13

image
Price is 0 and the reference empty. It could be updated.

  * Update supplier 2 for Product 13 with original values

image

  * Confirm state with GET

image

  * Delete the suppliers of the product 13

image

</details>

src/ApiPlatform/Resources/Product/VirtualProductFile.php
jolelievre: Good catch, even if the productId is not useful here, at least the REST convention is respected

src/ApiPlatform/Resources/Product/VirtualProductFile.php
jolelievre: Yep same format for everything, it may be worth checking this kind of url in the whole module We may have done similar mistakes in other domains

src/ApiPlatform/Resources/Product/ProductStock.php
jolelievre: It would have the same value true, but it's kind of a brain twist to understand it, no? 😅
I think this one was copied from how it was built in the core https://github.com/PrestaShop/PrestaShop/blob/837c662f3d5d27ff57b9f670f3b7244ff2c08b42/src/PrestaShopBundle/Form/Admin/Type/DeltaQuantityType.php#L54

jolelievre:
Regarding the `virtual-files` plural suggestion, I don't mind

Actually, I force the singular form on purpose (even by adding an exception in the rector rule) because today a product can have only one virtual product, and I wanted to emphasize this

But it also makes sense to favor respecting the plural form for convention (even if there's only one). I'll lean to the plural form for another reason though: if in the future we improve the virtual product and allow it to have multiple virtual files, at least the API URLs will natively be able to handle them

For the other singular exceptions, `stock` and `default-supplier` I think those two still make sense even in the future. The default supplier, without a doubt, there will always be only one default supplier among the list For the stock I wonder if a product could have multiple stocks, but since we removed the notion of advanced stock management and warehouses, I don't think so

I'm waiting for @ibahloul-ps feedback, and then I'll handle them all in one go

jolelievre:
@ibahloul-ps @Quetzacoalt91 latest commits:

- Virtual file URIs are plural: `POST /products/{productId}/virtual-files`, `POST /products/{productId}/virtual-files/{virtualProductFileId}`, `DELETE /products/virtual-files/{virtualProductFileId}`. PR description updated.
- `file` is no longer documented in the responses, only as the multipart binary part.
- `expirationDate`, `accessDays` and `downloadTimesLimit` are returned as `null` instead of being omitted.
- Shop images: both operations document that they always cover every shop. The query and command take a product id and no shop constraint, so `shopId`, `shopGroupId`, `shopIds` and `allShops` never restrict them.
- Attribute groups: `\d+` requirement plus the constraint exception mapped, so `0` returns 422 instead of 500.
- The custom serializers declare only the command they handle: the `'object'` / `'*'` entries were disabling the serializer support cache for every type.

Unchanged, on purpose:

- `productId` in the create body comes from the command schema, and the core `CQRSOpenApiFactoryTest` asserts it for this kind of multipart body.
- 200 on the update: the file already exists.
- `productId` ignored on the update: it builds the response, and no query reads a virtual file by its own id.
- 200 on shop images for an unknown product: needs a guard in `GetShopProductImagesHandler`.

Question: the DELETE is the only one without `productId`, since the create needs it for the command and the update for the response. Add it back for symmetry even though it would be decorative there?

src/ApiPlatform/Resources/Product/ProductStock.php
Quetzacoalt91: It was already difficult to understand why the negative is used in only one case :grimacing: but fair enough

Quetzacoalt91 (dismissed):
> Question: the DELETE is the only one without productId, since the create needs it for the command and the update for the response. Add it back for symmetry even though it would be decorative there?

I'd prefer to see the product ID but that's entirely your call. :)

Quetzacoalt91:
Blocked by https://github.com/PrestaShop/PrestaShop/pull/42960

Quetzacoalt91 (approved):

PR #305, opened 2026-07-02
Add regenerate-thumbnails endpoint

PrestaEdit:
| Questions         | Answers
| ----------------- | -------------------------------------------------------------
| Branch?           | dev
| Description?       | Adds `PUT /image-types/regenerate-thumbnails` (`RegenerateThumbnailsCommand`, scope `image_type_write`): regenerate the thumbnails of an image domain (`all`, `categories`, `manufacturers`, `suppliers`, `products`, `stores`) for a single image type (`imageTypeId`) or all of them (`0`), optionally erasing previous images.
| Type?             | improvement
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630
| How to test?      | `PUT /image-types/regenerate-thumbnails` with `{ "image": "suppliers", "imageTypeId": 0, "erasePreviousImages": false }` (client granted `image_type_write`) returns 204. Covered by `RegenerateThumbnailsEndpointTest`.
| Possible impacts? | Regenerates thumbnail images on disk for the selected domain.

PR #277, opened 2026-06-25
Add delete-images-from-type endpoint

PrestaEdit:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | Add the delete-images-from-type endpoint
| Type?             | new feature
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630
| Sponsor company   |

Adds **`DELETE /image-types/{imageTypeId}/images`** — `DeleteImagesFromTypeCommand`: removes the
generated thumbnail files for a given image type (`204`).

- `404` if the image type does not exist (`ImageTypeNotFoundException`)

Complements the ImageType resource. The integration test calls it on a fixture image type and asserts
a `204`. Reuses the `image_type_write` scope.

PR #232, opened 2026-06-15
Add ImageType resource

PrestaEdit:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | Add the ImageType resource (CRUD) to the Admin API
| Type?             | new feature
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630
| Sponsor company   |

Adds the Admin API endpoints for **image types** (Image Settings):

- `POST /image-types` — `AddImageTypeCommand`
- `GET /image-types/{imageTypeId}` — `GetImageTypeForEditing`
- `PATCH /image-types/{imageTypeId}` — `EditImageTypeCommand`
- `DELETE /image-types/{imageTypeId}` — `DeleteImageTypeCommand`
- `DELETE /image-types/bulk-delete` — `BulkDeleteImageTypeCommand`

An image type has a `name`, `width`/`height` and the entity flags it applies to
(`products`, `categories`, `manufacturers`, `suppliers`, `stores`). All fields are scalar
and match the CQRS command/query names, so no mapping is needed.

Adds an integration test covering the full CRUD + bulk delete, and the
`image_type_read` / `image_type_write` scopes.

PR #421, opened 2026-08-20
Merge the ImageType endpoints into one domain PR

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?      | Consolidates the three pending ImageType PRs into one domain PR, with every test fixture built through the API
| BC breaks?        | no
| Deprecations?     | no

## What this PR does

Merges #232, #277 and #305 into one PR, following the mutualisation done on the Product domain in #410.

### Endpoints

| GET | `/image-types/{imageTypeId}` | `GetImageTypeForEditing` | `image_type_read` |
| POST | `/image-types` | `AddImageTypeCommand` | `image_type_write` |
| PATCH | `/image-types/{imageTypeId}` | `EditImageTypeCommand` | `image_type_write` |
| DELETE | `/image-types/{imageTypeId}` | `DeleteImageTypeCommand` | `image_type_write` |
| DELETE | `/image-types/bulk-delete` | `BulkDeleteImageTypeCommand` | `image_type_write` |
| DELETE | `/image-types/{imageTypeId}/images` | `DeleteImagesFromTypeCommand` | `image_type_write` |
| PUT | `/image-types/regenerate-thumbnails` | `RegenerateThumbnailsCommand` | `image_type_write` |

### Tests

The three test classes become one `ImageTypeEndpointTest`.

**The delete-images test no longer works on someone else's data.** #277 had no way to create an image type, so it took whatever came first:

```php
self::$imageTypeId = (int) \Db::getInstance()->getValue(
    'SELECT `id_image_type` FROM `' . _DB_PREFIX_ . 'image_type` ORDER BY `id_image_type` ASC'
);
```

and then deleted that type's generated images. #232 has `POST /image-types`, so the test now creates the type it is about to wipe. That removes the raw `SELECT`, stops the test from touching a fixture image type, and makes it possible to assert something meaningful afterwards — that wiping the images leaves the type itself in place, which the original could not check without risking a false positive on a shared row.

The two absorbed classes also had no table reset of their own; they now inherit the `image_type` restore from the merged class.

### Version gate

The `#305` commit is titled *"Add regenerate-thumbnails endpoint (draft, 9.1+)"*. That note is stale — `RegenerateThumbnailsCommand` exposes the same constructor and the same getters on `9.0.3`, `9.1.4`, `9.2.x` and `develop`, so no `minVersion` is declared.

## How to test

```
POST   /image-types                      -> 200
GET    /image-types/{id}                 -> 200
PATCH  /image-types/{id}                 -> 200
DELETE /image-types/{id}                 -> 204, then the GET answers 404
DELETE /image-types/bulk-delete          -> 204, then each GET answers 404
DELETE /image-types/{id}/images          -> 204, and the type is still there
PUT    /image-types/regenerate-thumbnails -> 204
```

Covered by `ImageTypeEndpointTest`.

## Supersedes

- #232
- #277
- #305


mattgoud (changes requested):
Reviewed against a local install with the branch mounted and every endpoint called, so the codes below are observed rather than deduced. The happy paths are in good shape: GET, DELETE and the 404 all behave, the `Create` validation group works, and the router registers the seven URIs exactly as declared. Two problems on the write side, though.

## Blocking

### 1. `PATCH` has no validation at all, and it persists invalid data

```
POST  /image-types          {"name":"","width":-5,...}   422   name: This value should not be blank.
                                                               width: This value should be positive.

PATCH /image-types/{id}     {"name":""}                  200
PATCH /image-types/{id}     {"width":0}                  200
PATCH /image-types/{id}     {"width":-5}                 500
```

After the two accepted calls, the resource is persisted in a state the `POST` refuses:

```json
{"imageTypeId":17,"name":"","width":0,"height":90,...}
```

And the third answers a 500 that leaks the storage layer to the caller:

```
SQLSTATE[22003]: Numeric value out of range: 1264 Out of range value for column 'width' at row 1
```

The cause is in two halves, and fixing only one changes nothing. `CQRSPartialUpdate` carries no `validationContext`, **and** every constraint on the DTO is declared in the `Create` group alone:

```php
#[Assert\NotBlank(groups: ['Create'])]
public string $name;

#[Assert\NotNull(groups: ['Create'])]
#[Assert\Positive(groups: ['Create'])]
public int $width;
```

So adding `validationContext: ['groups' => ['Default', 'Update']]` on its own would still validate nothing, since no constraint belongs to `Default` or to `Update`. The constraints have to join the `Update` group as well, `groups: ['Create', 'Update']`, or move out of `Create` for the ones that must always hold.

`EditImageTypeCommand` does not guard these either, its setters take the value as is, so the API layer is the only place this can be caught.

### 2. An unknown image domain answers 500

```
PUT /image-types/regenerate-thumbnails   {"image":"not_a_domain","imageTypeId":0,"erasePreviousImages":false}
500   "not_a_domain" is not a valid backing value for enum "...\ImageSettings\ImageDomain"   (ValueError)
```

`RegenerateThumbnailsHandler` opens with `ImageDomain::from($command->getImage())`, and a backed enum throws `\ValueError` on an unknown value. That is not a domain exception, so no `exceptionToStatus` entry can catch it, and the two entries already declared on the resource never come into play for this case.

An `Assert\Choice` on `$image` fed from `ImageDomain::cases()` turns it into the 422 it should be, and doubles as documentation of the accepted values, which currently live only in a docblock.

## Worth fixing, not blocking

**`RegenerateThumbnails.php` has no `declare(strict_types=1)`.** It is the only one of the four new files without it. CI does not catch it today, which is probably why it slipped through.

**The `POST` answers only the identifier.**

```
POST /image-types   ->   201   {"imageTypeId":17}
```

The `PATCH` already replays `GetImageTypeForEditing` and answers the full resource. Adding the same `CQRSQuery` to `CQRSCreate` would make the create consistent with it and spare the caller a follow-up GET.

## Tests

There is no `testInvalid*` in the class. A single one asserting that `PATCH {"name":""}` answers 422 would have caught point 1 before this review.

`testEditImageType` checks `width`, `categories` and `height`, three of the eight properties. The other three tests of the class do assert the whole response, which is the bar set in ps_apiresources#438, so this one is the odd one out rather than the rule.

Three of its points did not survive a check, flagging them so they do not cost you a round trip.

- **`regenerate-thumbnails` does not need to be in `ApiResourceUriTemplateRector::SKIPPED_KEYWORDS`.** The router registers `/image-types/regenerate-thumbnails` exactly as declared and the Rector job is green. The segment is already plural, so the inflector has nothing to do to it.
- **The field alignment is fine**, so no `QUERY_MAPPING` or `CQRSCommandMapping` is required: the GET returns the eight declared properties populated, and a PATCH round-trips through them.
- Its reading of the missing `validationContext` stops one step short. On its own that attribute would not have changed any of the three calls in point 1 above.

One thing it did not raise and that I checked in case it looked like an omission: `EditableImageType` exposes `imageFitment` and `EditImageTypeCommand` can set it, but that field does not exist in 9.0.3, which this module still supports. Leaving it out of the resource is the right call.

kpodemski:
@mattgoud 

If pre-review doesn't do something what it should, or if it produces false-postives, let's make sure to improve it :)

PR #332, opened 2026-07-07
Add shop logos read endpoint

PrestaEdit:
| Questions         | Answers
| ----------------- | -------------------------------------------------------------
| Branch?           | dev
| Description?       | Exposes `GetLogosPaths` as `GET /shops/logos` (scope `shop_read`): returns the paths of the header, mail, invoice logos and favicon.
| Type?             | improvement
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630
| How to test?      | `GET /shops/logos` (client granted `shop_read`) returns `{ headerLogoPath, mailLogoPath, invoiceLogoPath, faviconPath }`. Covered by `ShopLogosEndpointTest`.
| Possible impacts? | None — read-only endpoint on a new resource.

PR #301, opened 2026-06-30
Add shop logos upload endpoint

PrestaEdit:
| Questions         | Answers
| ----------------- | -------------------------------------------------------------
| Branch?           | dev
| Description?       | Adds an endpoint to upload the shop logos: `PUT /shops/logos`. It accepts a `multipart/form-data` body with optional `uploadedHeaderLogo`, `uploadedMailLogo`, `uploadedInvoiceLogo` and `uploadedFavicon` files, and updates the corresponding shop logos. This exposes the `UploadLogosCommand`, previously unexposed. Scope `shop_write`.
| Type?             | improvement
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630
| How to test?      | `PUT /shops/logos` with a `multipart/form-data` body containing one or more of `uploadedHeaderLogo`, `uploadedMailLogo`, `uploadedInvoiceLogo`, `uploadedFavicon` (with a client granted `shop_write`); the matching logos are replaced. Unsupported extensions return 422. Covered by `UploadShopLogosEndpointTest`.
| Possible impacts? | Writes the uploaded logo images to the shop image directory.

PR #422, opened 2026-08-20
Merge the Shop endpoints into one domain PR

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?      | Consolidates the three pending Shop PRs into one domain PR, merging the two classes that declared the same /shops/logos URI
| BC breaks?        | no
| Deprecations?     | no

## What this PR does

Merges #231, #301 and #332 into one PR, following the mutualisation done on the Product domain in #410.

### Endpoints

| GET | `/shops/search` | `SearchShops` | `shop_read` |
| GET | `/shops/logos` | `GetLogosPaths` | `shop_read` |
| PUT | `/shops/logos` | `UploadLogosCommand` | `shop_write` |

### Design decisions

**Same URI, two classes — again.** As in the ApiClient secrets, #301 and #332 both declared `/shops/logos`: `UploadShopLogos` for the `PUT` and `ShopLogos` for the `GET`. They are now a single `ShopLogos` resource.

**Read and write are not the same shape here**, which is what made the split tempting: the write side takes four uploaded files, the read side returns the four paths the core stored them at. Rather than two classes, the resource declares both sides on one class with explicit directions:

- `uploadedHeaderLogo`, `uploadedMailLogo`, `uploadedInvoiceLogo`, `uploadedFavicon` → `#[ApiProperty(readable: false)]`, so they stay out of the responses
- the four `*Path` properties stay plain, which is what the read side of #332 already was

**The upload replays `GetLogosPaths` and answers `200` with the resulting paths**, so callers do not have to issue a second `GET` to read them back. I first marked the path properties `writable: false` to document the direction, but CI showed that flag also keeps them out of the *responses* — `GET /shops/logos` came back empty. It is not used that way anywhere in this module: the two places on `dev` that use it (`AttributeList`, `FeatureValueList`) pair it with `readable: false` to hide a property entirely. The path properties therefore stay plain, and the read side of #332 keeps the exact contract it shipped.

`FoundShop` (`/shops/search`) has nothing to do with the logos and is untouched.

### Tests

`ShopLogosEndpointTest` replaces `ShopLogosEndpointTest` + `UploadShopLogosEndpointTest`.

The upload still cannot be exercised with real files: the legacy uploader relies on `move_uploaded_file()` / `is_uploaded_file()`, which only succeed for genuine HTTP uploads and cannot run against a simulated request in the test kernel — the same limitation the Title image upload has. The endpoint wiring (routing, scope, multipart negotiation, command dispatch) is exercised with an empty payload.

What the merge adds is the assertion that was impossible while the two lived apart: **an upload that changed nothing must leave the logo paths exactly as the GET reported them.** The read test also pins the complete key set, so the write-only `uploaded*` properties failing to stay out of the response would fail the test.

## How to test

```
GET /shops/search?searchTerm=...  -> 200
GET /shops/logos                  -> 200 {headerLogoPath, mailLogoPath, invoiceLogoPath, faviconPath}
PUT /shops/logos (multipart)      -> 200 {headerLogoPath, mailLogoPath, invoiceLogoPath, faviconPath}
```

Covered by `ShopEndpointTest` and `ShopLogosEndpointTest`.

## Supersedes

- #231
- #301
- #332


mattgoud:
Heads up: I opened PrestaShop/ps_apiresources#438 to write down a testing convention that was never documented anywhere, and this PR is one of the two where it is not met yet.

The tests here assert individual fields but never the complete response, so the shape of the payload is not pinned: a renamed, removed or unexpectedly added field would go through green. The issue has the idiom and a couple of examples taken from the existing suite.

Nothing else is blocking on my side, and this is not a reproach on the PR: several of your other PRs already apply it, which is why the issue cites them as the target.

mattgoud:
Correction on my comment above, and my apologies: it is wrong for half of this PR.

I had only read `ShopEndpointTest.php`. `ShopLogosEndpointTest.php` does pin the response, and does it well:

```php
$this->assertEquals(
    ['headerLogoPath', 'mailLogoPath', 'invoiceLogoPath', 'faviconPath'],
    array_keys($logos)
);
```

Asserting the exact key set is the right tool here, and the comment above it says why: it is what proves the write-only `uploaded*` properties are never normalized back out. `testUploadLogosAnswersWithTheLogoPaths` then compares the whole payload twice. Nothing to change there.

What still stands is only `ShopEndpointTest::testSearchShops`, where the search result rows are checked with `assertArrayHasKey` on `id`, `color` and `name`, so an extra or renamed field on those rows would pass unnoticed.

Sorry for the noise, that was my mistake for reviewing on a partial read.

mattgoud (changes requested):
Reviewed against a local install with the branch mounted, so everything below is reproduced rather than read off the diff. Two blocking points, one documentation mismatch, two test gaps.

## Blocking

### 1. An empty `searchTerm` answers 500

```
GET /shops/search?searchTerm=        HTTP 500
GET /shops/search?searchTerm=%20     HTTP 500
GET /shops/search?searchTerm=shop    HTTP 200
```

Those are the real status lines, not the body of the debug payload. The core constructor asserts the term is not empty and throws:

```php
// src/Core/Domain/Shop/Query/SearchShops.php
throw new SearchShopException('Search term cannot be empty.');
```

`FoundShop` declares no `exceptionToStatus`, so it surfaces unmapped. The `required: true` parameter only guards the absent case, which correctly answers 422, and an empty value walks straight past it.

Mapping `SearchShopException` to 422 next to the existing parameter declaration is enough, plus a test covering the empty value.

### 2. The resource reserves a route it does not implement

`FoundShop` carries `#[ApiProperty(identifier: true)]` while declaring a single collection operation, so API Platform registers an item route for it:

```
_api_/found_shops/{id}{._format}_get   GET   /found_shops/{id}.{_format}
_controller: api_platform.action.not_exposed()
```

```
GET /admin-api/found_shops/1
404 {"detail":"This route does not aim to be called.","class":"ApiPlatform\\Metadata\\Exception\\NotExposedHttpException"}
```

That URI becomes unavailable to anyone, for a route that answers nothing. This is the same thing PrestaShop/PrestaShop#42852 is removing from the core's `Language` resource right now.

The fix is not the same as on the core, and I checked both. Dropping the attribute is not enough here, because the property is literally named `id` and API Platform picks it up as the identifier by convention. What works is:

```php
#[ApiProperty(identifier: false)]
public int $id;
```

With that, the router keeps only `_api_/shops/search_get_collection`, `GET /shops/search?searchTerm=shop` returns the identical payload, and `/found_shops/1` becomes an honest "No route found" 404.

## The docblock contradicts the code

`ShopLogos` opens with:

> The upload answers an empty 204 rather than replaying GetLogosPaths: the result of an upload is read back with the GET on the same URI.

Twelve lines below, the operation says the opposite, and so does the behaviour:

```
PUT /shops/logos   ->   200 {"headerLogoPath":"\/img\/logo.png","mailLogoPath":"\/img\/logo.png","invoiceLogoPath":"\/img\/logo.png","faviconPath":"\/img\/favicon.ico"}
```

The commit `Make the logo upload answer with the resulting paths` is what made it stale. The 200 is the better behaviour, only the class docblock and the `The upload keeps answering 204` paragraph of the PR description need to follow.

## Tests

`testSearchShops` asserts that `id`, `color` and `name` are present on every row. The endpoint actually returns two different shapes, and the distinction is the substance of the contract:

```
searchTerm=shop      [{"id":1,"color":"","name":"PrestaShop","groupId":1,"groupName":"Default","groupColor":""}]
searchTerm=Default   [{"id":1,"color":"","name":"Default"}]
```

Neither shape is pinned and `groupId`, `groupName` and `groupColor` are never exercised, although a default install produces both rows as above. Asserting the two payloads whole would cover the row types and the field set in one go.

The empty `searchTerm` of point 1 needs a test too.

Since it is sitting above and it reports a hard blocker, four of its points do not survive a check against the repository, and I would rather say so than let them cost you a round trip.

- **`?File` is not forbidden.** `tests/PHPStan/ApiResourcePropertyTypeRule.php` allows it explicitly: `ALLOWED_CLASS_SHORT_NAMES = ['DecimalNumber', 'DateTimeImmutable', 'File']`. That is why CI is green, and nothing needs changing. The report is reading `CONTEXT.md`, which still documents two exceptions and has fallen behind the rule. I will open a separate issue for that, it is a repository problem and not yours.
- **`$id` renamed to `$shopId`** would be wrong here. The collection mixes shop rows and shop-group rows, so on a group row that value is a shop-group id. With `identifier: false` above, the identifier naming rule no longer applies anyway.
- **The four path properties do not risk an uninitialized access.** `GetLogosPathsHandler` always builds `LogosPaths` with the four values, each falling back to the header logo path when the specific one is unset.
- **The result fields are already aligned**, so no result-side `QUERY_MAPPING` is needed: the live response carries all six properties populated.

The rest of the report is fair, and the `exceptionToStatus` line is the one that turned out to matter.

mattgoud:
Thanks, I re-ran the four points on a local install with the branch mounted and all four hold:

```
GET /shops/search?searchTerm=       422   (was 500)
GET /shops/search?searchTerm=%20    422   (was 500)
GET /found_shops/1                  404   and the route is gone from the router entirely
GET /shops/search?searchTerm=shop   200   payload unchanged
GET /shops/search?searchTerm=Default 200  payload unchanged
GET  /shops/logos                   200   unchanged
PUT  /shops/logos                   200   unchanged
```

`Integration tests (8.1, 9.2.x)` is red though, and it is my advice that broke it, so here is the diagnosis rather than just the red cross.

```
1) ShopEndpointTest::testSearchShopsReturnsAShopRow
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
         'id' => 1
         'color' => ''
-        'name' => 'PrestaShop'
+        'name' => 'test_shop'
         'groupId' => 1
```

The shop name is fixture data. It is `PrestaShop` on a plain install and `test_shop` on the CI one, so the only unstable value in the payload is that one field. Everything else in both rows matched, including `groupName`, the two ids and the empty colours, which is a good sign for the assertion itself.

One trap if you reach for the obvious source: **`Configuration::get('PS_SHOP_NAME')` is empty here**, I checked on the running shop. The name lives on the shop row itself, so it has to come from the object:

```php
$shopName = (new \Shop(1))->name;

$this->assertEquals(
    [[
        'id' => 1,
        'color' => '',
        'name' => $shopName,
        'groupId' => 1,
        'groupName' => 'Default',
        'groupColor' => '',
    ]],
    $this->getItem('/shops/search?searchTerm=shop', ['shop_read'])
);
```

`ApiTestCase` already instantiates `\Shop` directly, so the class is available in that context. Keeping `searchTerm=shop` as the query works on both installs, since it is a substring of `PrestaShop` and of `test_shop` alike.

`testSearchShopsReturnsAShopGroupRow` needs nothing: `Default` is the group name on both.

PR #376, opened 2026-07-13
Add GetCurrencyExchangeRate Admin API endpoint

PrestaEdit:
| Questions         | Answers                                                     |
|-------------------|-------------------------------------------------------------|
| Branch?           | dev                                                          |
| Description?      | Adds `GET /currencies/exchange-rates?isoCode=…` using `CQRSGet` with `GetCurrencyExchangeRate`. The handler returns an `ExchangeRate` value object (mono-arg VO wrapping a float) — `ValueObjectNormalizer` collapses it to scalar, and the `[_queryResult]` mapping surfaces it under the `exchangeRate` property. The `DecimalNumber`-typed resource prop matches the `ApiResourceScalarTypesRector` convention. |
| Type?             | new feature                                                  |
| Category?         | CO                                                           |
| BC breaks?        | no                                                           |
| Deprecations?     | no                                                           |
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630                       |
| How to test?      | Run `composer phpunit-integration`; `CurrencyExchangeRateEndpointTest` fetches USD and expects a numeric `exchangeRate` string. |

Retry of the previously-closed #346 with the correct understanding of the ExchangeRate VO shape (float-wrapping, not DecimalNumber).

PR #365, opened 2026-07-13
Add RefreshExchangeRates Admin API endpoint

PrestaEdit:
| Questions         | Answers                                                     |
|-------------------|-------------------------------------------------------------|
| Branch?           | dev                                                          |
| Description?      | Adds `POST /currencies/exchange-rate-refreshes` using `CQRSCreate` with `RefreshExchangeRatesCommand` (empty body). Delegates to `Currency::refreshCurrencies()` which fetches the PrestaShop currency XML feed. |
| Type?             | new feature                                                  |
| Category?         | CO                                                           |
| BC breaks?        | no                                                           |
| Deprecations?     | no                                                           |
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630                       |
| How to test?      | Run `composer phpunit-integration`; new `ExchangeRateRefreshEndpointTest` covers scope protection only — a full happy-path test would hit an external endpoint (`PS_CURRENCY_FEED_URL`) which is flaky in CI. `CannotRefreshExchangeRatesException` maps to HTTP 422 when the feed is unreachable. |

PR #360, opened 2026-07-13
Add EditUnofficialCurrency Admin API endpoint

PrestaEdit:
| Questions         | Answers                                                     |
|-------------------|-------------------------------------------------------------|
| Branch?           | dev                                                          |
| Description?      | Adds `PATCH /currencies/unofficials/{currencyId}` using `CQRSPartialUpdate` with `EditUnofficialCurrencyCommand`. All fields optional — ID-only ctor + setters pattern inherited from `AbstractEditCurrencyCommand`. Complements #356 (`POST /currencies/unofficials`). |
| Type?             | new feature                                                  |
| Category?         | CO                                                           |
| BC breaks?        | no                                                           |
| Deprecations?     | no                                                           |
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630                       |
| How to test?      | Run `composer phpunit-integration`; new `EditUnofficialCurrencyEndpointTest` seeds an unofficial currency, PATCHes the isoCode, verifies via DB row read, covers unknown-id → 404. |

Standalone file to avoid touching the in-progress `Currency.php` of PR #235.

PR #356, opened 2026-07-12
Add AddUnofficialCurrency Admin API endpoint

PrestaEdit:
| Questions         | Answers                                                     |
|-------------------|-------------------------------------------------------------|
| Branch?           | dev                                                          |
| Description?      | Adds `POST /currencies/unofficials` using `CQRSCreate` with `AddUnofficialCurrencyCommand`. Body: `{isoCode, exchangeRate, enabled}`. Response echoes the input plus the generated `currencyId` — same no-CQRSQuery echo pattern as `Country`/`Tax`. Complements PR #235 (which covers `AddCurrencyCommand` for official currencies) by adding the sibling unofficial-currency path (both extend `AbstractAddCurrencyCommand` in core). |
| Type?             | new feature                                                  |
| Category?         | CO                                                           |
| BC breaks?        | no                                                           |
| Deprecations?     | no                                                           |
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630                       |
| How to test?      | Run `composer phpunit-integration`; new `UnofficialCurrencyEndpointTest` covers scope protection and POST happy path. |

Standalone file (`UnofficialCurrency.php`) to avoid touching the in-progress `Currency.php` of PR #235.

PR #423, opened 2026-08-20
Merge the Currency endpoints into one domain PR

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?      | Consolidates the six pending Currency PRs into one domain PR, collapsing eight resource classes into five and making every write return the entity
| BC breaks?        | no
| Deprecations?     | no

## What this PR does

Merges #235, #344, #356, #360, #365 and #376 into one PR, following the mutualisation done on the Product domain in #410.

### Endpoints

| GET | `/currencies/{currencyId}` | `GetCurrencyForEditing` | `currency_read` |
| POST | `/currencies` | `AddCurrencyCommand` + `GetCurrencyForEditing` | `currency_write` |
| PATCH | `/currencies/{currencyId}` | `EditCurrencyCommand` + `GetCurrencyForEditing` | `currency_write` |
| DELETE | `/currencies/{currencyId}` | `DeleteCurrencyCommand` | `currency_write` |
| PUT | `/currencies/{currencyId}/toggle-status` | `ToggleCurrencyStatusCommand` + `GetCurrencyForEditing` | `currency_write` |
| POST | `/currencies/unofficials` | `AddUnofficialCurrencyCommand` + `GetCurrencyForEditing` | `currency_write` |
| PATCH | `/currencies/unofficials/{currencyId}` | `EditUnofficialCurrencyCommand` + `GetCurrencyForEditing` | `currency_write` |
| PUT | `/currencies/bulk-toggle-status` | `BulkToggleCurrenciesStatusCommand` | `currency_write` |
| DELETE | `/currencies/bulk-delete` | `BulkDeleteCurrenciesCommand` | `currency_write` |
| GET | `/currencies/references` | `GetReferenceCurrency` | `currency_read` |
| GET | `/currencies/exchange-rates` | `GetCurrencyExchangeRate` | `currency_read` |
| PUT | `/currencies/exchange-rates` | `RefreshExchangeRatesCommand` | `currency_write` |

**Eight resource classes become five**, six test classes become three.

### Design decisions

**The unofficial currency is not a different structure.** `AddUnofficialCurrencyCommand` and `EditUnofficialCurrencyCommand` extend `AbstractAddCurrencyCommand` / `AbstractEditCurrencyCommand`, exactly like the official ones — the difference is that the core does not resolve an unofficial currency against the CLDR reference data, not that it looks different. #356 and #360 each added a class describing the same fields as `Currency`. Their operations move onto the `Currency` resource and both classes are gone, which is the "one class per structure, several operations per class" rule from #410.

**The domain now speaks one language.** #360 exposed `localizedNames` / `localizedSymbols` / `localizedTransformations`; #235 exposed `names` / `symbols` / `transformations` and mapped them through `COMMAND_MAPPING`. Same commands underneath, two different public field names in the same domain. Merged onto the #235 naming.

**The toggle returns the entity too.** `PUT /currencies/{id}/toggle-status` answered `204` while every read on the same resource returned the currency; it now replays `GetCurrencyForEditing` like the create and the two updates.

**`POST /currencies` now returns the entity.** It declared no `CQRSQuery`, so it answered `{currencyId}` alone — while the `PATCH` immediately below it returned the whole currency. Both now replay `GetCurrencyForEditing`, as do the two unofficial operations.

**`exchange-rate-refreshes` is gone.** #365 declared `POST /currencies/exchange-rate-refreshes` — a resource name invented so that `ApiResourceUriTemplateRector` would accept a pluralized last segment. Refreshing the rates is an update of the very thing #376 reads, so it becomes `PUT /currencies/exchange-rates` on the `CurrencyExchangeRate` resource. Same URI, two operations, no invented noun.

### Tests

| Was | Now |
|-----|-----|
| `SELECT iso_code FROM ps_currency WHERE id_currency = ...` after the unofficial create | the create response |
| a raw `INSERT INTO ps_currency` to seed the fixture of the unofficial edit | `POST /currencies/unofficials`, which is in this PR |

The unofficial edit could only seed with SQL because the endpoint that creates one lived in a different PR — the same pattern as the Employee and ImageType domains.

### Two bugs the consolidation surfaced

**`GET /currencies/exchange-rates` answered with the ISO code alone.** `GetCurrencyExchangeRateHandler` returns a `QueryResult\ExchangeRate`. It is neither a scalar nor a `ValueObject` — `ValueObjectNormalizer::isValueObject()` only claims classes whose name contains `ValueObject` — so `QueryResultSerializerTrait` does not wrap it behind `_queryResult`, and it is normalized through its `getValue()` getter as `[value]`. #376's `['[_queryResult]' => '[exchangeRate]']` mapping could never fire. The mapping is now `['[value]' => '[exchangeRate]']`, and the test pins the complete key set.

**Every endpoint addressing an unofficial currency by id answers 500, and this one needs a core fix.** An unofficial currency has no numeric ISO code — that is what makes it unofficial — so `numeric_iso_code` is `NULL`. `CurrencyContextListener` reads `currencyId` with `$request->get()`, which also looks in the route attributes, so the currency named in the URI becomes the context currency; building it then fails, because `CurrencyContext::__construct()` types `$numericIsoCode` as a non-nullable `string`:

```
TypeError: CurrencyContext::__construct(): Argument #5 ($numericIsoCode)
must be of type string, null given, called in CurrencyContextBuilder.php on line 53
```

`POST /currencies/unofficials` is unaffected, because it carries no id. `GET /currencies/{id}`, `PATCH /currencies/unofficials/{id}` and the delete are all unusable on an unofficial currency until either `CurrencyContext` accepts a null numeric ISO code or the listener stops reading route attributes. The create test asserts the create response only, the edit test is skipped, and both carry the explanation in the test class.

**One gap is documented rather than papered over:** #365 shipped with no test method at all, only its `getProtectedEndpoints()` entry. `PUT /currencies/exchange-rates` is still not behaviourally tested, because `RefreshExchangeRatesCommand` calls the remote PrestaShop currency feed and a live assertion would fail with `CannotRefreshExchangeRatesException` whenever the runner has no network. The reason is now written down in the test class. The `GET` depends on that same feed, which is why no "unknown iso code returns 404" case is asserted either.

## How to test

```
POST   /currencies                          -> 201, the whole entity, identical to the GET
GET    /currencies/{id}                     -> 200
PATCH  /currencies/{id}                     -> 200, the updated entity
PUT    /currencies/{id}/toggle-status       -> 200, the currency it flipped, identical to the GET
POST   /currencies/unofficials              -> 201, the whole entity, unofficial: true
PATCH  /currencies/unofficials/{id}         -> 500 today, see the core bug above
PUT    /currencies/bulk-toggle-status       -> 204, then each GET shows the new status
DELETE /currencies/bulk-delete              -> 204
GET    /currencies/references               -> 200
GET    /currencies/exchange-rates?isoCode=USD -> 200 {isoCode, exchangeRate}
PUT    /currencies/exchange-rates           -> 204   (calls the remote currency feed)
```

Covered by `CurrencyEndpointTest`, `CurrencyExchangeRateEndpointTest` and `ReferenceCurrencyEndpointTest`.

## Supersedes

- #235
- #344
- #356
- #360
- #365
- #376

All six will be closed once the CI is green here.


mattgoud (changes requested):
Reviewed against a local install with the branch mounted and every endpoint called. Two things first, since they are the ones I have been finding elsewhere today and this PR gets both right: the router registers the twelve URIs exactly as declared, with **no phantom item route** anywhere, and the `POST` answers the full resource rather than just the identifier. The `shopIds` round trip works too, more on that below.

Three problems on the write side, all of them the same shape: an operation accepts a field it cannot honour and answers as if it had.

## Blocking

### 1. `PATCH /currencies/{currencyId}` accepts `isoCode` and drops it

```
GET   /currencies/2                       isoCode = CAD
PATCH /currencies/2  {"isoCode":"CHF"}    200
GET   /currencies/2                       isoCode = CAD
```

`EditCurrencyCommand` has no `setIsoCode`. Its sibling `EditUnofficialCurrencyCommand` does, which is why the unofficial `PATCH` can change it and the official one cannot. Both operations hang off the same `Currency` class, so the official `PATCH` advertises `isoCode` in its schema, accepts it, answers 200 and silently ignores it.

The same call with an empty string behaves identically:

```
PATCH /currencies/2  {"isoCode":""}       200   and the value is untouched
```

A caller has no way to tell the difference between a field that was applied and one that was discarded.

### 2. `PUT /currencies/{currencyId}/toggle-status` ignores its body

```
enabled = false
PUT /currencies/2/toggle-status  {"enabled":false}   200   ->  enabled = true
PUT /currencies/2/toggle-status  {"enabled":false}   200   ->  enabled = false
```

`ToggleCurrencyStatusCommand` takes the id and nothing else, so it flips. The operation is declared with `allowEmptyBody: true` on the `Currency` resource, which still advertises the whole `Currency` schema as the input, `enabled` included. A client sending `enabled: false` to make sure a currency is off will turn it on one time in two.

The behaviour is the core's and it is not for this PR to change, but the resource should stop presenting a body it discards.

### 3. The two bulk endpoints disagree on an unknown id

```
DELETE /currencies/bulk-delete        {"currencyIds":[999999]}   500
PUT    /currencies/bulk-toggle-status {"currencyIds":[999999]}   204
```

The 500 is `BulkDeleteCurrenciesException`, `"Failed to delete all of selected currencies"`. Neither bulk class declares `exceptionToStatus`, so nothing maps it. The automated report raised this and it is real, with the exact class named above. The toggle answering 204 for the same input is the other half of the problem: one of the two is wrong, and they should at least agree.

## Worth a look

`exchangeRate` sent as a JSON string is rejected:

```
POST /currencies  {"exchangeRate":"1.5", ...}   400   must be one of "float" ("string" given)
POST /currencies  {"exchangeRate":1.5,   ...}   201
```

The property is typed `DecimalNumber`, which the conventions call for precisely so that a decimal does not go through a float. The only accepted wire format being a JSON number puts it back through one. Worth at least documenting that the string form is not accepted.

Its issue 3 is the one that mattered and is in point 3 above. Three of the others I would set aside.

- **`shopIds` missing from `COMMAND_MAPPING` is a false positive.** `AbstractAddCurrencyCommand` and `AbstractEditCurrencyCommand` both expose `setShopIds(array $shopIds)`, so the property maps by name and needs no entry. The asymmetry the report noticed is real and is exactly why `QUERY_MAPPING` carries `associatedShopIds => shopIds` while `COMMAND_MAPPING` does not. Verified end to end:

```
POST /currencies {"isoCode":"CAD","exchangeRate":1.5,"enabled":true,"shopIds":[1]}
201 {"currencyId":2, ..., "shopIds":[1]}
```

- **The missing `validationContext` on the two `CQRSPartialUpdate` operations** is narrower than described. `PATCH {"exchangeRate":-5}` already answers 422 today. The gap is worth closing for the future, it is not currently letting anything through.
- **`ReferenceCurrency.$isoCode` not marked `identifier: true`** is a cosmetic difference. I checked the router with both resources in place and neither declaration produces an extra route here, since both already carry an item operation on a fixed URI. Consistency is a fair request, nothing depends on it.

## Tests

`testAddCurrency` and `testToggleCurrencyStatus` compare one endpoint's response to another's:

```php
$this->assertEquals($this->getItem('/currencies/' . $currencyId, ['currency_read']), $this->lastCreatedCurrency);
```

That pins the two against each other but never pins the field set, so a property lost on both sides at once stays green. See PrestaShop/ps_apiresources#438. One expected array written out in full, as `testGetImageType` does in the ImageType PR, would cover it.

Nothing covers points 1, 2 and 3 above either, and each is a two-line test.

PR #313, opened 2026-07-06
Add cart update-product-quantity endpoint

PrestaEdit:
| Questions         | Answers
| ----------------- | -------------------------------------------------------------
| Branch?           | dev
| Description?       | Fills the `UpdateProductQuantityInCartCommand` gap: `PUT /carts/{cartId}/products/{productId}` (scope `cart_write`) to set the quantity of a product in a cart.
| Type?             | improvement
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630
| How to test?      | `PUT /carts/{id}/products/{productId}` with `{ "quantity": 3 }` (client granted `cart_write`) sets the product quantity in the cart. Covered by `CartProductQuantityEndpointTest`.
| Possible impacts? | None beyond the cart targeted by the request.

PR #310, opened 2026-07-03
Add cart add-product endpoint

PrestaEdit:
| Questions         | Answers
| ----------------- | -------------------------------------------------------------
| Branch?           | dev
| Description?       | Fills the `AddProductToCartCommand` gap: `PUT /carts/{cartId}/products` (scope `cart_write`) to add a product (optionally a specific combination) with a quantity to a cart.
| Type?             | improvement
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630
| How to test?      | `PUT /carts/{id}/products` with `{ "productId": x, "quantity": 2 }` (client granted `cart_write`) adds the product to the cart. A missing cart/product returns 404. Covered by `CartProductEndpointTest`.
| Possible impacts? | None beyond the cart targeted by the request.

PR #424, opened 2026-08-20
Merge the Cart endpoints into one domain PR

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?      | Consolidates the six pending Cart PRs into one domain PR, merging the three cart-line classes and moving every assertion onto the cart details endpoint
| BC breaks?        | no
| Deprecations?     | no

## What this PR does

Merges #299, #309, #310, #312, #313 and #318 into one PR, following the mutualisation done on the Product domain in #410.

### Endpoints

| GET | `/carts/{cartId}/details` | `GetCartForViewing` | `cart_read` |
| PUT | `/carts/{cartId}/products` | `AddProductToCartCommand` | `cart_write` |
| PUT | `/carts/{cartId}/products/{productId}` | `UpdateProductQuantityInCartCommand` | `cart_write` |
| DELETE | `/carts/{cartId}/products/{productId}` | `RemoveProductFromCartCommand` | `cart_write` |
| PUT | `/carts/{cartId}/currencies` | `UpdateCartCurrencyCommand` | `cart_write` |
| PUT | `/carts/{cartId}/languages` | `UpdateCartLanguageCommand` | `cart_write` |
| PUT | `/carts/{cartId}/process-order-emails` | `SendProcessOrderEmailCommand` | `order_write` |

**Six resource classes become four, six test classes become one.**

### Design decisions

**Three classes for one cart line.** `CartProduct` (add), `CartProductQuantity` (update quantity) and `CartProductRemoval` (remove) declared identical properties, and the last two declared the *same URI* — `/carts/{cartId}/products/{productId}` — from two different resource classes. They are now one `CartProduct` with three operations, using explicit `uriVariables` so the add operation binds only `cartId`.

**The cart line deliberately returns 204.** There is no query to replay: the cart line has no read side of its own, and `GetCartForViewing` builds the whole `CartDetails` view rather than a single line. Returning that view from a `CartProduct` operation would mean denormalizing it into a class that does not have its fields. The result of a write is observed through `GET /carts/{cartId}/details`, which is what the tests now do.

### Tests

Six test classes become one `CartEndpointTest`, and this is where the per-endpoint split had really accumulated:

- **`createCart()` existed in five copies** — one per test class, with small differences between them. It is now a single documented helper.
- **The removal test seeded its own cart line with SQL:**
  ```php
  // Seed the cart line directly so the test does not depend on the add-product flow.
  \Db::getInstance()->insert('cart_product', [...]);
  ```
  That reasoning made sense for a standalone PR and makes none for a domain PR. Add → change quantity → remove is now one chained lifecycle test going through the API at every step.
- **Assertions moved off raw SQL and off the legacy object:**

| Was | Now |
|-----|-----|
| `SELECT quantity FROM ps_cart_product WHERE ...` | the `cart_quantity` of `cartSummary.products` in the details |
| `SELECT COUNT(*) FROM ps_cart_product WHERE ...` after removal | the same products array, now empty |
| `(new \Cart($cartId))->id_currency` | `cartCurrencyId` in the details |

**On fixtures:** carts are created by the front office and by the order-creation flow — the Cart domain exposes no "add cart" command, so `createCart()` still uses the legacy object. Two SQL *lookups* also remain and are marked as such in the code: finding an active, in-stock, combination-less product (the products listing cannot express that filter), and finding a second installed currency (the Currency domain has no listing endpoint on `dev`). Neither creates anything.

`testUpdateCartLanguage` only asserts the command is accepted: `GetCartForViewing` does not expose the cart language, so that write has no read side to check against.

## One question for the reviewers

`PUT /carts/{cartId}/process-order-emails` declares the `order_write` scope, while every other `/carts/...` endpoint — including the already merged `CartEmail` (#314) — uses `cart_write`. #318 is approved as it stands so the scope is left untouched here, but if that was an oversight rather than a deliberate choice, this is the PR to fix it in.

## How to test

```
GET    /carts/{id}/details                -> 200 {cartId, cartCurrencyId, customerInformation, orderInformation, cartSummary}
PUT    /carts/{id}/products                -> 204, then the details list the product
PUT    /carts/{id}/products/{productId}    -> 204, then the details show the new quantity
DELETE /carts/{id}/products/{productId}    -> 204, then the details list no product
PUT    /carts/{id}/currencies              -> 204, then the details show the new cartCurrencyId
PUT    /carts/{id}/languages               -> 204
PUT    /carts/{id}/process-order-emails    -> 204   (422 on cartId=0)
```

Covered by `CartEndpointTest`.

## Supersedes

- #299
- #309
- #310
- #312
- #313
- #318

All six will be closed once the CI is green here.


src/ApiPlatform/Resources/Cart/CartDetails.php
```diff
@@ -0,0 +1,62 @@
...
+use ApiPlatform\Metadata\ApiResource;
+use PrestaShop\PrestaShop\Core\Domain\Cart\Exception\CartNotFoundException;
+use PrestaShop\PrestaShop\Core\Domain\Cart\Query\GetCartForViewing;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSGet;
+use Symfony\Component\HttpFoundation\Response;
+
+#[ApiResource(
+    operations: [
+        new CQRSGet(
+            uriTemplate: '/carts/{cartId}/details',
```
Quetzacoalt91, line 33: This should be 
```suggestion
            uriTemplate: '/carts/{cartId}',
```

as on the other domains

src/ApiPlatform/Resources/Cart/CartDetails.php
```diff
@@ -0,0 +1,62 @@
...
+
+    public array $customerInformation;
+
+    public array $orderInformation;
+
+    public array $cartSummary;
+
+    public const QUERY_MAPPING = [
+        '[cartId]' => '[cartId]',
+    ];
```
Quetzacoalt91, line 61: Can be removed as the mapping has the same origin and destination properties

src/ApiPlatform/Resources/Cart/CartDetails.php
```diff
@@ -0,0 +1,62 @@
...
+            requirements: ['cartId' => '\d+'],
+            CQRSQuery: GetCartForViewing::class,
+            scopes: [
+                'cart_read',
+            ],
+            CQRSQueryMapping: self::QUERY_MAPPING,
+        ),
+    ],
+    exceptionToStatus: [
+        CartNotFoundException::class => Response::HTTP_NOT_FOUND,
```
Quetzacoalt91, line 43: To be added: CartConstraintException

tests/Integration/ApiPlatform/CartEndpointTest.php
```diff
@@ -0,0 +1,236 @@
...
+    {
+        $cartId = $this->createCart();
+        $productId = $this->getSimpleInStockProductId();
+
+        $this->updateItem(
+            '/carts/' . $cartId . '/products',
+            ['productId' => $productId, 'quantity' => 2],
+            ['cart_write'],
+            Response::HTTP_NO_CONTENT
+        );
```
Quetzacoalt91, line 145: Can you make this call twice (eventually with different quantities, and update the expected result? It is supposed to show the quantity is added.

tests/Integration/ApiPlatform/CartEndpointTest.php
```diff
@@ -0,0 +1,236 @@
...
+            array_keys($cart)
+        );
+        $this->assertSame($cartId, $cart['cartId']);
+        $this->assertSame([], $cart['cartSummary']['products']);
+    }
+
+    public function testGetNonExistentCartDetails(): void
+    {
+        $this->requestApi('GET', '/carts/999999/details', null, ['cart_read'], Response::HTTP_NOT_FOUND);
+    }
```
Quetzacoalt91, line 128: Can you add a test with "/carts/0/details" ?

Quetzacoalt91 (changes requested):

PR #303, opened 2026-07-01
Add Category bulk enable/disable endpoints

PrestaEdit:
| Questions         | Answers
| ----------------- | -------------------------------------------------------------
| Branch?           | dev
| Description?       | Fills the two Category status commands not covered by the existing category endpoints: `PUT /categories/bulk-enable` (`BulkEnableCategoriesCommand`) and `PUT /categories/bulk-disable` (`BulkDisableCategoriesCommand`). Both scoped `category_write` and take a `categoryIds` array.
| Type?             | improvement
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630
| How to test?      | `PUT /categories/bulk-enable` and `PUT /categories/bulk-disable` with `{ "categoryIds": [id1, id2] }` (client granted `category_write`) enable/disable the given categories. A missing category returns 404. Covered by `CategoryBulkStatusEndpointTest`.
| Possible impacts? | None beyond the categories targeted by each request.

PR #425, opened 2026-08-20
Merge the Category endpoints into one domain PR

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?      | Consolidates the pending Category PRs into one domain PR, and drops a bulk-status endpoint pair that duplicates what dev already has
| BC breaks?        | no
| Deprecations?     | no

## What this PR does

Merges #214, #300, #303 and #395 into one PR, following the mutualisation done on the Product domain in #410.

### Endpoints added

| GET | `/categories/{categoryId}/status` | `GetCategoryIsEnabled` | `category_read` |
| GET | `/categories/trees` | `GetCategoriesTree` | `category_read` |
| PUT | `/categories/update-positions` | `UpdateCategoryPositionCommand` | `category_write` |

### #303 is dropped, not merged

`dev` already exposes `PUT /categories/bulk-update-status`, which takes `{categoryIds, enabled}` and toggles the set either way. #303 added `PUT /categories/bulk-enable` and `PUT /categories/bulk-disable` on top of it — three URIs for one capability, and `bulk-update-status` is the shape every other domain uses (employees, stores, zones, tax rules groups).

`BulkEnableCategoriesCommand` and `BulkDisableCategoriesCommand` are added to `EXCLUDED_CQRS_CLASSES` with the `USELESS_DUPLICATE` reason, so the tracking table records them as intentionally not exposed rather than as a gap. This is the mechanism #410 used for `GetProductIsEnabled`, `GetAssociatedSuppliers` and the others.

Its test class goes with it — and it is a good illustration of the cost of the split: it seeded categories with `new \Category()->add()` and asserted with `(new \Category($id))->active`, while `CategoryEndpointTest`, in the very same domain, was already creating them through `POST /categories` and asserting through the `GET`.

### On the position endpoint

`UpdateCategoryPosition` uses a plain `CQRSUpdate` and exposes the drag-and-drop payload of the admin grid (`positions` as `prefix_{parentId}_{categoryId}` tokens, plus `way` and `foundFirst`), rather than the `UpdatePosition` metadata that `AttributePosition` and `AttributeGroupPosition` use. That is not an oversight: `UpdatePosition` needs a `positionDefinition` grid service, and there is no `prestashop.core.grid.category.position_definition` in the core — only attribute, attribute_group, carrier, cms_page, cms_page_category, feature, feature_value, image and product have one. `UpdateCategoryPositionCommand` is the only way in, and it takes that payload.

### Tests

One `CategoryEndpointTest`. The tree test is folded in and strengthened: it creates a category through `POST /categories` and asserts that it appears in the flattened tree, where the standalone test could only check that `$tree[0]` had the expected keys — it had no category of its own to look for.

## How to test

```
GET /categories/{id}/status     -> 200
GET /categories/trees           -> 200, the created category is in the tree
PUT /categories/update-positions -> 204, then the GETs show the swapped positions
```

Covered by `CategoryEndpointTest`.

## Supersedes

- #214
- #300
- #395

and closes #303 as a duplicate (no endpoint from it is kept).


boo-code:
`UpdateCategoryPosition` carries `way` and `foundFirst` over from #395 unchanged, and with them the sibling-ordering corruption measured on that PR: with `foundFirst: true`, `UpdateCategoryPositionHandler` skips `Category::cleanPositions()`, and on `9.1.x` with three siblings, moving the first one "up" to position 2 went

```
positions before   0, 1, 2
positions after    1, 2, 2     duplicate at 2, nothing at 0
```

in both `category` and `category_shop`. `#[Assert\NotNull] public bool $foundFirst` makes sending it mandatory, so a client reaches that path on the default contract rather than by doing something unusual.

PrestaShop/PrestaShop#42216 re-indexes unconditionally and removes it whatever the flag says. It is still open.

#432 carries a "PATCH and DELETE need a core change" section naming PrestaShop/PrestaShop#41742 for the same kind of dependency. This one has the same shape and no equivalent line, which is the only thing worth changing here - the payload discussion above already settles why the endpoint takes `way`/`foundFirst` at all.

boo-code:
Two things about the position side of this PR.

**No endpoint here returns a position, so the last How to test line cannot be followed.** For `PUT /categories/update-positions -> 204, then the GETs show the swapped positions`: `CategoryTree` exposes `categoryId`, `enabled`, `name`, `displayName` and `children`, and core's `CategoryForTree` has no position getter either. The `Category` resource's `public int $position` is removed in this same PR, and `GetCategoryForEditingHandler` never populated it - `EditableCategory` has no `getPosition()` and the handler's constructor call passes none. Removing the property is therefore right, it could never be filled; but with it gone there is no read path left and the swap cannot be observed through any GET in this PR.

That leaves the domain able to write positions and not read them. `AttributeGroup` went the other way in PrestaShop/PrestaShop#41844, which returns the position from the for-editing query. The same change on `GetCategoryForEditing` would make the line above testable and give `update-positions` something to assert against.

Minor, same property: #404 adds an `ApiProperty` description to `Category::$position`. If this merges first that hunk has nothing left to document - worth knowing now rather than in a rebase.

src/ApiPlatform/Resources/Category/CategoryTree.php
```diff
@@ -0,0 +1,61 @@
...
+            CQRSQuery: GetCategoriesTree::class,
+            scopes: [
+                'category_read',
+            ],
+            CQRSQueryMapping: [
+                '[_context][langId]' => '[languageId]',
+                '[_context][shopId]' => '[shopId]',
+            ],
+            ApiResourceMapping: [
+                '[active]' => '[enabled]',
```
Quetzacoalt91, line 41: Unfortunately, this applies only to the first level of the array

image

src/ApiPlatform/Resources/Category/UpdateCategoryPosition.php
```diff
@@ -0,0 +1,78 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Category\Exception\CategoryNotFoundException;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSUpdate;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Validator\Constraints as Assert;
+
+#[ApiResource(
+    operations: [
+        new CQRSUpdate(
+            uriTemplate: '/categories/update-positions',
+            output: false,
```
Quetzacoalt91, line 38: The lack of Query after the update looks acceptable to me, as there is no read equivalent of this command.

tests/Integration/ApiPlatform/CategoryEndpointTest.php
```diff
@@ -314,4 +424,42 @@ private function createTemporaryCategories(): array
...
+            'parentCategoryId' => 2,
+            'shopIds' => [1],
+        ], ['category_write'])['categoryId'];
+
+        $tree = $this->getItem('/categories/trees', ['category_read']);
+        $this->assertNotEmpty($tree);
+        $this->assertEquals(
+            ['categoryId', 'enabled', 'name', 'displayName', 'children'],
+            array_keys($tree[0])
+        );
```
Quetzacoalt91, line 448: You're only checking the first element of the array. If you made a deep check, you'd have found an issue on the `disabled` field.

tests/Integration/ApiPlatform/CategoryEndpointTest.php
```diff
@@ -314,4 +424,42 @@ private function createTemporaryCategories(): array
 
         return [$cat1['categoryId'], $cat2['categoryId']];
     }
+
+    /**
+     * The tree is the read side of the whole category hierarchy, so it is asserted against a
+     * category this suite created rather than against whatever the fixtures ship: the
+     * standalone test could only check that the first row had the expected keys.
```
Quetzacoalt91, line 431: The point of this test is to check the whole structure of the response is correct. If you stop at the first row, you basically do the same check as `GET /categories/{categoryId}`. Please add assertions of all elements, including the children, and check the values as well.

tests/Integration/ApiPlatform/CategoryEndpointTest.php
```diff
@@ -220,6 +232,35 @@ public function testUpdateCategoryStatus(): void
...
+            Request::METHOD_PATCH,
+            '/categories/3/status',
+            ['enabled' => true],
+            ['category_write'],
+            Response::HTTP_OK
+        );
+
+        $status = $this->getItem('/categories/3/status', ['category_read']);
+        $this->assertEquals(3, $status['categoryId']);
+        $this->assertTrue($status['enabled']);
```
Quetzacoalt91, line 261: Please make a check of the whole array, as in https://github.com/PrestaShop/ps_apiresources/blob/dcf05ff910c6473f0c98801a02becc891177d6ab/tests/Integration/ApiPlatform/DiscountTypeEndpointTest.php#L68-L83

Quetzacoalt91 (changes requested):

PR #359, opened 2026-07-13
Add EditLanguage Admin API endpoint

PrestaEdit:
| Questions         | Answers                                                     |
|-------------------|-------------------------------------------------------------|
| Branch?           | dev                                                          |
| Description?      | Adds `PATCH /languages/{languageId}` using `CQRSPartialUpdate` with `EditLanguageCommand`. All fields are optional — `EditLanguageCommand` uses the ID-only ctor + setters pattern. Complements #351 (`GET /languages/{languageId}/details`) and #357 (`POST /languages`). |
| Type?             | new feature                                                  |
| Category?         | CO                                                           |
| BC breaks?        | no                                                           |
| Deprecations?     | no                                                           |
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630                       |
| How to test?      | Run `composer phpunit-integration`; new `EditLanguageEndpointTest` covers scope protection, rename happy path, and unknown-id → 404. |

Standalone file (`EditLanguage.php`) to avoid touching the in-progress `Language.php` of PR #255 (Language delete + status endpoints).

PR #357, opened 2026-07-12
Add AddLanguage Admin API endpoint

PrestaEdit:
| Questions         | Answers                                                     |
|-------------------|-------------------------------------------------------------|
| Branch?           | dev                                                          |
| Description?      | Adds `POST /languages` using `CQRSCreate` with `AddLanguageCommand`. Body: `name`, `isoCode` (2-letter), `tagIETF`, `shortDateFormat`, `fullDateFormat`, `rtl`, `enabled`, `shopIds[]`, plus optional `flagImagePath` / `noPictureImagePath` (default empty — the handler no-ops the image copy when the paths are empty). Response echoes the input plus the generated `languageId` (no `CQRSQuery` read-back — Country/Tax echo pattern). Complements PR #351 (`GET /languages/{languageId}/details`). |
| Type?             | new feature                                                  |
| Category?         | CO                                                           |
| BC breaks?        | no                                                           |
| Deprecations?     | no                                                           |
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630                       |
| How to test?      | Run `composer phpunit-integration`; new `AddLanguageEndpointTest` covers scope protection and POST happy path. |

PR #351, opened 2026-07-11
Add GetLanguageForEditing Admin API endpoint

PrestaEdit:
| Questions         | Answers                                                     |
|-------------------|-------------------------------------------------------------|
| Branch?           | dev                                                          |
| Description?      | Adds \`GET /languages/{languageId}\` using \`CQRSGet\` with \`GetLanguageForEditing\` — a mono-argument Query returning the editable Language shape. Includes integration tests (scope protection, happy path with type assertions on all 10 fields, unknown-id → 404). |
| Type?             | new feature                                                  |
| Category?         | CO                                                           |
| BC breaks?        | no                                                           |
| Deprecations?     | no                                                           |
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630                       |
| How to test?      | Run \`composer phpunit-integration\`; new \`LanguageEndpointTest\` covers scope protection, GET happy path, unknown-id 404. |

Endpoint added:
- \`GET /languages/{languageId}\` — CQRS mapping \`shopAssociation → shopIds\`, \`active → enabled\`; \`is\`-prefix and boolean-field Rector conventions applied.

Note: PR #178 previously bundled Language read + write; it was closed. This scope-limited GET picks up the read half only, following the pattern of the recently landed simple read endpoints.

PR #255, opened 2026-06-22
Add Language status & delete endpoints

PrestaEdit:
| Questions         | Answers                                                       |
|-------------------|---------------------------------------------------------------|
| Description?      | Adds the Language status and deletion endpoints to the Admin API. The upload-heavy create/edit operations (flag & no-picture images) are intentionally left out of this PR. |
| Type?             | new feature                                                   |
| BC breaks?        | no                                                            |
| Deprecations?     | no                                                            |
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630                         |
| Sponsor company   |                                                               |
| How to test?      | Run `tests/Integration/ApiPlatform/LanguageEndpointTest.php`, or call the new endpoints with a `language_write` scoped API client. |

### Endpoints added

- `PATCH /languages/{languageId}/set-status` — `{ enabled }` (`ToggleLanguageStatusCommand`, 204)
- `PUT /languages/bulk-update-status` — `{ languageIds, enabled }` (`BulkToggleLanguagesStatusCommand`, 204)
- `DELETE /languages/{languageId}` — (`DeleteLanguageCommand`, 204)
- `DELETE /languages/bulk-delete` — `{ languageIds }` (`BulkDeleteLanguagesCommand`, 204)

All operations use the `language_write` scope. The default language cannot be deleted nor disabled (mapped to `422`), and a missing language returns `404`.

PR #427, opened 2026-08-20
Merge the Language endpoints into one domain PR

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?      | Consolidates the four pending Language PRs, which were the same resource cut in four conflicting pieces, into a single class
| BC breaks?        | no
| Deprecations?     | no

## What this PR does

Merges #255, #351, #357 and #359 into one PR, following the mutualisation done on the Product domain in #410.

### Endpoints

| GET | `/languages/{languageId}` | `GetLanguageForEditing` | `language_read` |
| POST | `/languages` | `AddLanguageCommand` + `GetLanguageForEditing` | `language_write` |
| PATCH | `/languages/{languageId}` | `EditLanguageCommand` + `GetLanguageForEditing` | `language_write` |
| PATCH | `/languages/{languageId}/set-status` | `ToggleLanguageStatusCommand` + `GetLanguageForEditing` | `language_write` |
| DELETE | `/languages/{languageId}` | `DeleteLanguageCommand` | `language_write` |
| PUT | `/languages/bulk-update-status` | `BulkToggleLanguagesStatusCommand` | `language_write` |
| DELETE | `/languages/bulk-delete` | `BulkDeleteLanguagesCommand` | `language_write` |

**Four resource classes become one.**

### These four PRs were one resource cut in four

`Language` (#255: set-status + delete), `LanguageForEditing` (#351: the GET), `Language` again (#357: the POST) and `EditLanguage` (#359: the PATCH). #255 and #351 both wrote `src/ApiPlatform/Resources/Language/Language.php` **and** `tests/Integration/ApiPlatform/LanguageEndpointTest.php`, with different content and different scopes (`language_write` vs `language_read`) — cherry-picking them in sequence conflicts. Whichever had been merged first, the other would have needed a rewrite.

**Read and write are not the same set of fields**, which is what made splitting them look reasonable. They are now one class where each field carries the direction it actually has:

- `locale` is `#[ApiProperty(writable: false)]` — the core derives it from the IETF tag and `AddLanguageCommand` has no locale argument;
- `flagImagePath` and `noPictureImagePath` are `#[ApiProperty(readable: false)]` — `GetLanguageForEditing` does not return them.

**The GET loses its `/details` suffix.** #351 declared `GET /languages/{languageId}/details` because `/languages/{languageId}` was already taken by another PR's `DELETE`. In one resource, the GET of an item is just the GET of the item.

**`POST` and `PATCH` return the entity**, both replaying `GetLanguageForEditing`. #357's create answered `{languageId, name, isoCode, enabled}` from the command result alone, and #359's update returned a partially-filled object.

### Tests

Three test classes become one `LanguageEndpointTest`, and every fixture goes through the API:

| Was | Now |
|-----|-----|
| `self::addLanguageByLocale('es-ES')` (command bus) to seed languages | `POST /languages`, which #357 provides |
| `(new \Language($id))->active` after set-status and bulk-status | `enabled` in the `GET` |
| `Validate::isLoadedObject(new \Language($id))` after delete and bulk-delete | the `GET` answering `404` |
| `SELECT name FROM ps_lang WHERE id_lang = ...` after the edit | the update response, asserted against the `GET` |

The edit test also stops operating on `$languageId = 1` — the shop's default language — and edits one it created itself.

## How to test

```
POST   /languages                     -> 201, the whole entity, no image paths in the response
GET    /languages/{id}                -> 200, identical to the POST response
PATCH  /languages/{id}                -> 200, the updated entity
PATCH  /languages/{id}/set-status     -> 200, the language, identical to the GET
PUT    /languages/bulk-update-status  -> 204, then each GET shows the new status
DELETE /languages/{id}                -> 204, then the GET answers 404
DELETE /languages/bulk-delete         -> 204, then each GET answers 404
```

Covered by `LanguageEndpointTest`.

## Supersedes

- #255
- #351
- #357
- #359


src/ApiPlatform/Resources/Language/Language.php
```diff
@@ -0,0 +1,164 @@
...
+ * as input, while the flag and no-picture image paths are inputs the editing query does not
+ * return. Both stay on this resource, with the direction they actually have.
+ */
+#[ApiResource(
+    operations: [
+        new CQRSGet(
+            // Not /languages/{languageId}: CQRSGet does not extend ApiPlatform\Metadata\Get,
+            // so ApiPlatform adds its own NotExposedOperation on the default item URI and that
+            // route answers 404 instead of this one.
+            uriTemplate: '/languages/{languageId}/details',
```
Quetzacoalt91: I'd prefer to follow the same pattern as for the other domains.
This route won't work for the already released versions of PrestaShop, so we will have to provide minimum version of PS.

Can you try the initial route, as you listed in the PR description, along https://github.com/PrestaShop/PrestaShop/pull/42852?

If this work, we can change the GET endpoint with:

```suggestion
            uriTemplate: '/languages/{languageId}',
            extraProperties: [
                'minVersion' => '9.3.0',
            ],
```

Quetzacoalt91 (changes requested):
Hi @PrestaEdit,
I haven't checked the whole PR yet,  but here's a first review I'd like you to have a look.

Thanks

PrestaEdit:
Applied, thanks — `/languages/{languageId}` + `extraProperties: ['minVersion' => '9.3.0']`.

Verified locally with #42852 applied on top of `develop`: the 15 integration tests in `LanguageEndpointTest` pass. Had to temporarily bump `minVersion` down to `9.2.0` during the run (since core still reports `9.2.0` and the version guard would otherwise strip the endpoint), then restored `9.3.0` before pushing. Tests updated to hit the new route as well.

src/ApiPlatform/Resources/Language/Language.php
```diff
@@ -0,0 +1,164 @@
...
+    #[Assert\NotBlank(groups: ['Create'])]
+    public string $fullDateFormat;
+
+    /**
+     * Write only in practice: GetLanguageForEditing does not return the image paths, so they
+     * are only ever set on the way in.
+     */
+    public string $flagImagePath = '';
+
+    public string $noPictureImagePath = '';
```
Quetzacoalt91, line 138:
```suggestion
    #[ApiProperty(readable: false, writable: true)]
    public string $flagImagePath = '';

    #[ApiProperty(readable: false, writable: true)]
    public string $noPictureImagePath = '';
```

tests/Integration/ApiPlatform/LanguageEndpointTest.php
```diff
@@ -0,0 +1,230 @@
...
+
+    public function testCreateLanguage(): int
+    {
+        $language = $this->createLanguage('ts');
+
+        $this->assertSame('Test TS', $language['name']);
+        $this->assertSame('ts', $language['isoCode']);
+        $this->assertTrue($language['enabled']);
+        // The create replays GetLanguageForEditing, so it answers exactly what the GET does
+        $this->assertEquals($this->getLanguage($language['languageId']), $language);
```
Quetzacoalt91, line 121: You testing the same call to the same function, but you don't check the whole array.

Please add an assertEquals that checks everything, as we can find on other files. For instance https://github.com/PrestaShop/ps_apiresources/blob/dcf05ff910c6473f0c98801a02becc891177d6ab/tests/Integration/ApiPlatform/DiscountTypeEndpointTest.php#L68-L83

Quetzacoalt91 (approved):

Quetzacoalt91 (changes requested):
Wrong option selected :sweat_smile:

src/ApiPlatform/Resources/Language/Language.php
```diff
@@ -0,0 +1,164 @@
...
+    #[Assert\NotBlank(groups: ['Create'])]
+    public string $shortDateFormat;
+
+    #[Assert\NotBlank(groups: ['Create'])]
+    public string $fullDateFormat;
+
+    /**
+     * Write only in practice: GetLanguageForEditing does not return the image paths, so they
+     * are only ever set on the way in.
+     */
```
Quetzacoalt91, line 135: Why not putting 
```suggestion
    #[ApiProperty(readable: false, writable: true)]
```

as described in your PR description?

mattgoud:
Heads up: I opened PrestaShop/ps_apiresources#438 to write down a testing convention that was never documented anywhere, and this PR is one of the two where it is not met yet.

The tests here assert individual fields but never the complete response, so the shape of the payload is not pinned: a renamed, removed or unexpectedly added field would go through green. The issue has the idiom and a couple of examples taken from the existing suite.

Nothing else is blocking on my side, and this is not a reproach on the PR: several of your other PRs already apply it, which is why the issue cites them as the target.

jolelievre:
@Quetzacoalt91 with the changes done on the core to unlock this one it means that several endpoints added here will only work in 9.3 (cf https://github.com/PrestaShop/PrestaShop/pull/42852)

Unless we add this in the 9.2.x branch, there would still be a limitation but to 9.2.1 maybe (depending on when the fix reaches the patch)

Quetzacoalt91:
> @Quetzacoalt91 with the changes done on the core to unlock this one it means that several endpoints added here will only work in 9.3 (cf [PrestaShop/PrestaShop#42852](https://github.com/PrestaShop/PrestaShop/pull/42852))
> 
> Unless we add this in the 9.2.x branch, there would still be a limitation but to 9.2.1 maybe (depending on when the fix reaches the patch)

That's not a regression of the 9.2.0, that's why I'm targeting `develop` for PS 9.3.0. It wouldn't be fair to force this change in a patch when we refuse other fixes.

PR #386, opened 2026-07-13
Add BulkDeleteProductsFromOrderReturn Admin API endpoint

PrestaEdit:
| Questions         | Answers                                                     |
|-------------------|-------------------------------------------------------------|
| Branch?           | dev                                                          |
| Description?      | Adds `DELETE /order-returns/{orderReturnId}/products/bulk-delete` using `CQRSDelete` with `BulkDeleteProductsFromOrderReturnCommand`. Body: `{stagedProductRows: [{order_detail_id, customization_id?}, ...]}` — snake_case matches the Command's ctor read shape (`$row['order_detail_id']`, `$row['customization_id']`). Complements #372 (single-product delete + list). |
| Type?             | new feature                                                  |
| Category?         | CO                                                           |
| BC breaks?        | no                                                           |
| Deprecations?     | no                                                           |
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630                       |
| How to test?      | Scope-protection test only — happy-path coverage would need seeding an OrderReturn with real OrderReturnDetail rows which the existing test infrastructure does not currently expose. |

**⚠️ Depends on PR #220** — the OrderReturn product-mgmt commands are 9.1+/develop. 9.0.3 CI matrix legs will fail until #220 (drop 9.0.3) lands.

PR #372, opened 2026-07-13
Add OrderReturn products Admin API endpoints (list + delete)

PrestaEdit:
| Questions         | Answers                                                     |
|-------------------|-------------------------------------------------------------|
| Branch?           | dev                                                          |
| Description?      | Adds:<br/>- `GET /order-returns/{orderReturnId}/products` — GetOrderReturnProducts, returns per-line rows `{orderDetailId, customizationId, reference, productName, quantity, customization}`<br/>- `DELETE /order-returns/{orderReturnId}/products/{orderDetailId}` — DeleteProductFromOrderReturnCommand, optional `?customizationId=` query parameter (defaults to 0 for regular product lines) |
| Type?             | new feature                                                  |
| Category?         | CO                                                           |
| BC breaks?        | no                                                           |
| Deprecations?     | no                                                           |
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630                       |
| How to test?      | Run `composer phpunit-integration`; `OrderReturnProductsEndpointTest` covers scope protection only. Happy-path coverage would need seeding a full OrderReturn + OrderReturnDetail row set which the existing test infrastructure does not currently expose. |

**⚠️ Depends on PR #220** — the core OrderReturn product-mgmt commands were introduced in PS 9.1+/develop and do not exist at the 9.0.3 tag. The 9.0.3 CI matrix legs will fail here until #220 (drop 9.0.3 from CI) lands.

PR #371, opened 2026-07-13
Add OrderReturn delete + bulk-delete Admin API endpoints

PrestaEdit:
| Questions         | Answers                                                     |
|-------------------|-------------------------------------------------------------|
| Branch?           | dev                                                          |
| Description?      | Adds:<br/>- `DELETE /order-returns/{orderReturnId}` (DeleteOrderReturnCommand)<br/>- `DELETE /order-returns/bulk-delete` (BulkDeleteOrderReturnsCommand)<br/>Single-delete op added to the existing `OrderReturn.php` (which already has GET + PATCH); bulk op lives in a new `BulkOrderReturns.php` file. |
| Type?             | new feature                                                  |
| Category?         | CO                                                           |
| BC breaks?        | no                                                           |
| Deprecations?     | no                                                           |
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630                       |
| How to test?      | Run `composer phpunit-integration`; `OrderReturnDeleteEndpointTest` seeds order-return rows via `\Db::insert('order_return', …)`, exercises DELETE + bulk-DELETE, verifies row-count = 0. |

**⚠️ Depends on PR #220** — `DeleteOrderReturnCommand` / `BulkDeleteOrderReturnsCommand` were introduced in PS 9.1+/develop and do not exist at the 9.0.3 tag. The 9.0.3 CI matrix legs will fail here until #220 (drop 9.0.3 from CI) lands.

PR #430, opened 2026-08-20
Merge the OrderReturn endpoints into one domain PR

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?      | Consolidates the three pending OrderReturn PRs into one domain PR, and gives two of them the behavioural tests they shipped without
| BC breaks?        | no
| Deprecations?     | no

## What this PR does

Merges #371, #372 and #386 into one PR, following the mutualisation done on the Product domain in #410.

### Endpoints

| DELETE | `/order-returns/{orderReturnId}` | `DeleteOrderReturnCommand` | `order_return_write` |
| DELETE | `/order-returns/bulk-delete` | `BulkDeleteOrderReturnsCommand` | `order_return_write` |
| GET | `/order-returns/{orderReturnId}/products` | `GetOrderReturnProducts` | `order_return_read` |
| DELETE | `/order-returns/{orderReturnId}/products/{orderDetailId}` | `DeleteProductFromOrderReturnCommand` | `order_return_write` |
| DELETE | `/order-returns/{orderReturnId}/products/bulk-delete` | `BulkDeleteProductsFromOrderReturnCommand` | `order_return_write` |

### Two of the three had no behavioural test

#372 and #386 shipped with a `getProtectedEndpoints()` declaration and nothing else — scope protection covered, behaviour not. That was hard to avoid one PR at a time: **#372's products listing is the only read side of #386's bulk delete**, and #371's delete had nothing to check against either, so it fell back to `SELECT COUNT(*) FROM ps_order_return`.

One `OrderReturnActionsEndpointTest` now covers the five endpoints:

| Endpoint | Asserted through |
|----------|------------------|
| products listing | the complete row structure and the seeded order detail ids |
| delete one product | the listing, which no longer contains it |
| bulk delete products | the listing, now empty |
| delete an order return | `GET /order-returns/{orderReturnId}` answering `404` |
| bulk delete order returns | the same `GET`, for each id |

**On fixtures:** order returns are created by the customer from the front office. The `OrderReturn` domain exposes only `DeleteOrderReturnCommand`, `BulkDeleteOrderReturnsCommand` and `UpdateOrderReturnStateCommand` — there is no way to create one through the Admin API, so the seeding stays SQL. It is now a single documented helper, and it also creates the `order_return_detail` rows the product endpoints need, which is what makes those tests possible at all.

## How to test

```
GET    /order-returns/{id}/products               -> 200, the returned lines
DELETE /order-returns/{id}/products/{detailId}    -> 204, then the listing drops it
DELETE /order-returns/{id}/products/bulk-delete   -> 204, then the listing is empty
DELETE /order-returns/{id}                        -> 204, then the GET answers 404
DELETE /order-returns/bulk-delete                 -> 204, then each GET answers 404
```

Covered by `OrderReturnActionsEndpointTest`.

## Supersedes

- #371
- #372
- #386


src/ApiPlatform/Resources/OrderReturn/BulkOrderReturns.php
```diff
@@ -0,0 +1,56 @@
...
+            uriTemplate: '/order-returns/bulk-delete',
+            extraProperties: [
+                'minVersion' => '9.2.0',
+            ],
+            CQRSCommand: BulkDeleteOrderReturnsCommand::class,
+            scopes: ['order_return_write'],
+        ),
+    ],
+    exceptionToStatus: [
+        OrderReturnNotFoundException::class => Response::HTTP_NOT_FOUND,
```
nicosomb, line 45: Only OrderReturnNotFoundException is mapped. BulkDeleteOrderReturnsCommand's constructor throws OrderReturnConstraintException for an invalid id (OrderReturnId::assertIsIntegerGreaterThanZero), unmapped here so it falls through to 500. Needs OrderReturnConstraintException mapped to 422.

src/ApiPlatform/Resources/OrderReturn/OrderReturnProduct.php
```diff
@@ -0,0 +1,73 @@
...
+        OrderReturnConstraintException::class => Response::HTTP_UNPROCESSABLE_ENTITY,
+        // A merchandise return must keep at least one product: a client error, not a 500
+        CannotDeleteLastProductFromOrderReturnException::class => Response::HTTP_UNPROCESSABLE_ENTITY,
+    ],
+)]
+class OrderReturnProduct
+{
+    public int $orderReturnId;
+
+    public int $orderDetailId;
```
nicosomb, line 70: No ApiProperty(identifier: true) on orderDetailId, unlike every other non-bulk single-target resource in the module (CategoryDelete, TaxRule, SearchEngine/DeleteSearchEngine, Address all mark theirs). Worth aligning for consistency.

src/ApiPlatform/Resources/OrderReturn/OrderReturnProductList.php
```diff
@@ -0,0 +1,59 @@
...
+
+    public int $customizationId;
+
+    public string $reference;
+
+    public string $productName;
+
+    public int $quantity;
+
+    public bool $customization;
```
nicosomb, line 58: GetOrderReturnProductsHandler builds each row from OrderReturnProductForEditing, which also exposes getCustomizationFields() (the actual customization content, e.g. text/file inputs), not exposed here. The DTO says a line is customized but never what the customization contains.

nicosomb (commented):
Two more things worth noting, not really actionable in this PR:

Bulk delete isn't atomic. AbstractBulkCommandHandler::handleBulkAction only catches exceptions matching the type passed to it (DeleteOrderReturnException here), but OrderReturnRepository::delete() throws OrderReturnNotFoundException on a missing id, a sibling exception, not a subtype. Hitting a missing id partway through a batch re-throws immediately and abandons the rest of the array, with no rollback of what was already deleted. This is shared Core behavior, not specific to this PR, and no test here exercises a mixed valid/invalid batch.

Two items from the automated pre-review don't hold up. allowEmptyBody: false being absent on BulkOrderReturns/BulkOrderReturnProducts has no effect, false is already the framework default (CQRSApiSerializer::isEmptyBodyAllowed). And OrderReturnProductList's field mapping is fine as is, every property, including the isCustomization getter mapping to customization, matches GetOrderReturnProducts's result by name.

PR #373, opened 2026-07-13
Add MailTemplate get + edit Admin API endpoints

PrestaEdit:
| Questions         | Answers                                                     |
|-------------------|-------------------------------------------------------------|
| Branch?           | dev                                                          |
| Description?      | Adds two endpoints on a single template file (identified by `templateName` + `locale` + `source` + optional `moduleName`):<br/>- `GET /mail-templates/{templateName}?locale=…&source=…&moduleName=…` (GetEmailBodyTemplateForEditing → returns `htmlContent` + `txtContent`)<br/>- `PATCH /mail-templates/{templateName}` body `{locale, source, moduleName?, htmlContent, txtContent}` (EditEmailBodyTemplateCommand)<br/>Follows #370 (list). |
| Type?             | new feature                                                  |
| Category?         | CO                                                           |
| BC breaks?        | no                                                           |
| Deprecations?     | no                                                           |
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630                       |
| How to test?      | Run `composer phpunit-integration`; scope-protection tests only — a full happy-path test would need to seed a writable per-locale mail-template file tree. |

**⚠️ Depends on PR #220** — the core MailTemplate domain was introduced in PS 9.1+/develop and does not exist at the 9.0.3 tag. The 9.0.3 CI matrix legs will fail here until #220 (drop 9.0.3 from CI) lands.

PR #370, opened 2026-07-13
Add MailTemplate listing Admin API endpoint

PrestaEdit:
| Questions         | Answers                                                     |
|-------------------|-------------------------------------------------------------|
| Branch?           | dev                                                          |
| Description?      | Adds `GET /mail-templates?locale=…` using `CQRSGetCollection` with `GetEmailBodyTemplatesForListing`. Items expose `templateName`, `source`, `moduleName`, `hasHtml`, `hasTxt` (filesystem scan of core + module mail directories for the requested locale). Read-only for now — `GetEmailBodyTemplateForEditing` / `EditEmailBodyTemplateCommand` need the composite `EmailTemplateSource` VO which is less API-friendly and left for a follow-up. |
| Type?             | new feature                                                  |
| Category?         | CO                                                           |
| BC breaks?        | no                                                           |
| Deprecations?     | no                                                           |
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630                       |
| How to test?      | Run `composer phpunit-integration`; `MailTemplateListEndpointTest` fetches the `en-US` locale, expects a non-empty list (PS ships core English templates). |

**⚠️ Depends on PR #220** — the `MailTemplate` domain was introduced in PS 9.1+/develop and does not exist at the 9.0.3 tag. The 9.0.3 CI matrix legs will fail here until #220 (drop 9.0.3 from CI) lands.

PR #223, opened 2026-06-11
Add MailTemplate generate endpoint

PrestaEdit:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | Add a MailTemplate generate endpoint to the Admin API
| Type?             | new feature
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630
| Sponsor company   |

Exposes `GenerateThemeMailTemplatesCommand` through **`PUT /mail-templates`**, which
(re)generates the mail templates of a theme for a given language. Body: `themeName`,
`language` (locale, e.g. `en`) and optional `overwriteTemplates`.

Adds an integration test and the `mail_template_write` scope.

PR #431, opened 2026-08-20
Merge the MailTemplate endpoints into one domain PR

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?      | Consolidates the three pending MailTemplate PRs into one domain PR, and gives the get/edit endpoints the behavioural test they shipped without
| BC breaks?        | no
| Deprecations?     | no

## What this PR does

Merges #223, #370 and #373 into one PR, following the mutualisation done on the Product domain in #410.

### Endpoints

| GET | `/mail-templates?locale=` | `GetEmailBodyTemplatesForListing` | `mail_template_read` |
| PUT | `/mail-templates` | `GenerateThemeMailTemplatesCommand` | `mail_template_write` |
| GET | `/mail-templates/{templateName}?locale=&source=` | `GetEmailBodyTemplateForEditing` | `mail_template_read` |
| PATCH | `/mail-templates/{templateName}` | `EditEmailBodyTemplateCommand` | `mail_template_write` |

### #373 had no behavioural test

It shipped with a `getProtectedEndpoints()` declaration and nothing else. On its own it could not do much better: fetching a mail template requires a template name that really exists on the installation, and the endpoint that tells you which ones do is #370's listing — in another PR.

The merged test chains them: it lists the templates for `en-US`, picks a `core` one from the listing, gets it, edits its HTML and text content, then reads the edit back through the get. Nothing assumes `order_conf` is installed, which is what the `getProtectedEndpoints()` entry had to hardcode.

### One expectation corrected

This is the one write in these PRs that cannot return the modified entity, and the reason is worth recording. `CommandProcessor::denormalizeCommandResult()` builds the `CQRSQuery` from the URI variables merged with the *command result*. `EditEmailBodyTemplateHandlerInterface::handle()` returns `void`, and `GetEmailBodyTemplateForEditing` requires `$locale` and `$source` — which only ever exist in the request body. Nothing carries them across, so the `PATCH` declares `output: false` and answers an empty `204`.

Returning the template would need either a handler that returns the edited `EditableEmailBodyTemplate`, or a way to feed the request body into the replayed query.

`EditEmailBodyTemplateCommand` also types `$moduleName` as a plain `string` with no default, so the request has to carry it even for a core template.

## How to test

```
GET   /mail-templates?locale=en                             -> 200, the template rows
GET   /mail-templates/{name}?locale=en&source=core          -> 200 {templateName, htmlContent, txtContent, ...}
PATCH /mail-templates/{name}                                -> 204, then the GET shows the new content
PUT   /mail-templates                                       -> 204
```

Covered by `MailTemplateEndpointTest`.

## Supersedes

- #223
- #370
- #373


src/ApiPlatform/Resources/MailTemplate/MailTemplate.php
```diff
@@ -0,0 +1,97 @@
...
+            scopes: ['mail_template_read'],
+            openapiContext: [
+                'parameters' => [
+                    ['name' => 'locale', 'in' => 'query', 'required' => true, 'schema' => ['type' => 'string', 'example' => 'en'], 'description' => 'Language ISO code, the mails/<iso> directory the template lives in'],
+                    ['name' => 'source', 'in' => 'query', 'required' => true, 'schema' => ['type' => 'string', 'enum' => ['core', 'module']]],
+                    ['name' => 'moduleName', 'in' => 'query', 'required' => false, 'schema' => ['type' => 'string', 'default' => '']],
+                ],
+            ],
+        ),
+        new CQRSPartialUpdate(
```
nicosomb, line 54: No validationContext on this operation, unlike the pattern used elsewhere in the module. Without it, only the Default group validates on PATCH.

src/ApiPlatform/Resources/MailTemplate/MailTemplate.php
```diff
@@ -0,0 +1,97 @@
...
+class MailTemplate
+{
+    #[ApiProperty(identifier: true)]
+    public string $templateName;
+
+    #[Assert\NotBlank(groups: ['Create'])]
+    public string $locale;
+
+    /** 'core' or 'module' */
+    #[Assert\Choice(choices: ['core', 'module'], groups: ['Create'])]
```
nicosomb, line 85: locale and source use groups: ['Create'], but there is no CQRSCreate operation on this resource and no validationContext activates that group anywhere. These constraints never fire. EditEmailBodyTemplateCommand's constructor requires locale and source with no default, so a PATCH missing either currently falls through to a raw serializer error instead of a clean 422.

src/ApiPlatform/Resources/MailTemplate/GenerateThemeMailTemplates.php
```diff
@@ -0,0 +1,52 @@
...
+
+#[ApiResource(
+    operations: [
+        new CQRSUpdate(
+            uriTemplate: '/mail-templates',
+            output: false,
+            CQRSCommand: GenerateThemeMailTemplatesCommand::class,
+            scopes: ['mail_template_write'],
+        ),
+    ],
```
nicosomb, line 38: No exceptionToStatus here. GenerateThemeMailTemplatesHandler throws InvalidArgumentException for an unknown language (line 68) and for an unknown themeName (FolderThemeCatalog::getByName, line 102), both unmapped, both easy to trigger from a bad request. Needs a mapping to 422.

src/ApiPlatform/Resources/MailTemplate/MailTemplateList.php
```diff
@@ -0,0 +1,76 @@
...
+            ApiResourceMapping: self::API_RESOURCE_MAPPING,
+            scopes: ['mail_template_read'],
+            parameters: new Parameters([
+                new QueryParameter(
+                    key: 'locale',
+                    required: true,
+                    description: 'Language ISO code, the mails/<iso> directory to list (e.g. en, fr)'
+                ),
+            ]),
+            openapiContext: [
```
nicosomb, line 50: locale is declared both here and in the parameters block above. Redundant, and the two descriptions can drift out of sync. The parameters declaration also gets runtime enforcement, this one doesn't.

nicosomb (commented):
One point from the automated pre-review doesn't hold up: item 4, $source missing from MailTemplateList::API_RESOURCE_MAPPING. EmailBodyTemplateRepository::scanDirectory() in Core already returns the raw key as 'source' (same name as the DTO property), so it maps by name without needing an entry, the same way shopIds did on the CmsPageCategory PR. No fix needed there.

tests/Integration/ApiPlatform/MailTemplateEndpointTest.php
```diff
@@ -0,0 +1,165 @@
...
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+declare(strict_types=1);
+
+namespace PsApiResourcesTest\Integration\ApiPlatform;
+
+use Symfony\Component\HttpFoundation\Response;
+
+class MailTemplateEndpointTest extends ApiTestCase
```
nicosomb, line 27: No testInvalid* method in this class, required by CONTEXT.md for validation error coverage (e.g. PATCH missing locale/source, PUT missing themeName, both expecting 422 with assertValidationErrors).

tests/Integration/ApiPlatform/MailTemplateEndpointTest.php
```diff
@@ -0,0 +1,165 @@
...
+
+        return $template;
+    }
+
+    /**
+     * @depends testGetMailTemplate
+     *
+     * @param array{templateName: string, source: string} $template
+     */
+    public function testEditMailTemplate(array $template): void
```
nicosomb, line 115: This overwrites the html/txt content of a real core mail template on disk and never restores it, there's no tearDownAfterClass here. Every run of this suite leaves that template's files permanently changed.

PR #432, opened 2026-08-20
Merge the SqlRequest endpoints into one domain PR

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?      | Consolidates the four pending SqlRequest PRs into one domain PR, drops a superseded duplicate and fixes a namespace-driven URI
| BC breaks?        | no
| Deprecations?     | no

## What this PR does

Merges #239, #240, #334 and #396 into one PR, following the mutualisation done on the Product domain in #410.

### Endpoints

| GET | `/sql-requests/{sqlRequestId}` | `GetSqlRequestForEditing` | `sql_management_read` |
| POST | `/sql-requests` | `AddSqlRequestCommand` + `GetSqlRequestForEditing` | `sql_management_write` |
| PATCH | `/sql-requests/{sqlRequestId}` | `EditSqlRequestCommand` + `GetSqlRequestForEditing` | `sql_management_write` |
| DELETE | `/sql-requests/{sqlRequestId}` | `DeleteSqlRequestCommand` | `sql_management_write` |
| DELETE | `/sql-requests/bulk-delete` | `BulkDeleteSqlRequestCommand` | `sql_management_write` |
| GET | `/sql-requests/{sqlRequestId}/execution-results` | `GetSqlRequestExecutionResult` | `sql_management_read` |
| GET | `/sql-requests/database-tables` | `GetDatabaseTablesList` | `sql_management_read` |
| GET | `/sql-requests/database-tables/{tableName}/fields` | `GetDatabaseTableFieldsList` | `sql_management_read` |

### #239 is superseded by #240

Both created `src/ApiPlatform/Resources/SqlRequest/SqlRequest.php`, `BulkDeleteSqlRequests.php` and `SqlRequestEndpointTest.php`. #240 declares the same create and get as #239, plus the partial update and the delete — a strict superset. #240's version is the one kept; there is nothing in #239 to graft back.

### #396's URI was driven by its namespace

It declared `GET /sql-management/{sqlRequestId}/execution-results`. The id in that URI is a **sql request** id, and `ApiResourceUriTemplateRector` derives the first segment from the resource namespace — so putting the class under `SqlManagement` forced a sql request to be addressed under another noun, exactly one level away from the `/sql-requests/{sqlRequestId}` its siblings use.

The class moves to the `SqlRequest` namespace and the URI becomes `/sql-requests/{sqlRequestId}/execution-results`. Its scope also becomes `sql_management_read` — it declared `sql_request_read`, which exists nowhere else in the domain.

### `POST /sql-requests` returns the entity

It declared no `CQRSQuery`, so it answered `{sqlRequestId}` alone while the `PATCH` right below it returned the whole request. Both now replay `GetSqlRequestForEditing`, and the test asserts the create response is identical to the following `GET`.

### PATCH and DELETE need a core change

`EditSqlRequestCommand::__construct()` and `DeleteSqlRequestCommand::__construct()` only accept a `SqlRequestId` value object, while the CQRS normalizer builds the command from the scalar id in the URI — so both requests fail with a `TypeError` before the handler is reached.

PrestaShop/PrestaShop#41742 accepts a scalar id in both constructors. `testEditSqlRequest` (and `testDeleteSqlRequest`, which depends on it) stay red here until that PR is merged, following the same convention as #410.

### Tests

The execution-result test creates its saved SQL request through `POST /sql-requests` instead of `INSERT INTO ps_request_sql` — the create endpoint is in this PR now. It still selects from a stable seed table so the assertions do not drift when the languages fixture changes, and it now pins the complete key set of the response and of each row.

## How to test

```
POST   /sql-requests                                    -> 201, the whole entity, identical to the GET
GET    /sql-requests/{id}                               -> 200
PATCH  /sql-requests/{id}                               -> 200, the updated entity
DELETE /sql-requests/{id}                               -> 204, then the GET answers 404
DELETE /sql-requests/bulk-delete                        -> 204, then each GET answers 404
GET    /sql-requests/{id}/execution-results             -> 200 {sqlRequestId, columns, rows}
GET    /sql-requests/database-tables                    -> 200 {tables}
GET    /sql-requests/database-tables/{table}/fields     -> 200 {tableName, fields}
```

Covered by `SqlRequestEndpointTest` and `DatabaseTablesEndpointTest`.

## Supersedes

- #240
- #334
- #396

and closes #239, whose resource #240 already contains in full.


boo-code:
The execution-result endpoint keeps the broad description ("Sensitive attributes (e.g. password fields) are automatically obfuscated"), which is what I suggested on #396 - depending on the core fix rather than restating a narrower guarantee in the endpoint text. That still holds, but the dependency it rests on is not named anywhere in this PR, and it has not merged.

Until PrestaShop/PrestaShop#42219 lands, `GetSqlRequestExecutionResultHandler::hideSensitiveData()` masks `RequestSql::$attributes`, which is two keys:

```php
public $attributes = [
    'passwd' => '*******************',
    'secure_key' => '*******************',
];
```

so a saved `SELECT` over `ps_api_client.client_secret`, `ps_customer_session.token` or `ps_employee_session.token` returns them verbatim - here as JSON over HTTP rather than a table an employee is reading in the back office.

This PR already states that kind of dependency for `EditSqlRequestCommand` / `DeleteSqlRequestCommand` and PrestaShop/PrestaShop#41742. #42219 has the same shape and no equivalent line.

PR #375, opened 2026-07-13
Add GetCmsPageCategoriesForBreadcrumb Admin API endpoint

PrestaEdit:
| Questions         | Answers                                                     |
|-------------------|-------------------------------------------------------------|
| Branch?           | dev                                                          |
| Description?      | Adds `GET /cms-page-categories/{currentCategoryId}/breadcrumb` using `CQRSGetCollection` with `GetCmsPageCategoriesForBreadcrumb`. The core `Breadcrumb` result is an `IteratorAggregate` of `BreadcrumbItem {cmsPageCategoryId, name}`; the CQRSApiNormalizer iterates it to a plain items list which `CQRSGetCollection` surfaces as an array. |
| Type?             | new feature                                                  |
| Category?         | CO                                                           |
| BC breaks?        | no                                                           |
| Deprecations?     | no                                                           |
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630                       |
| How to test?      | Run `composer phpunit-integration`; `CmsPageCategoryBreadcrumbEndpointTest` fetches the breadcrumb for ROOT (id=1) and verifies each row has `cmsPageCategoryId` + `name`. |

PR #362, opened 2026-07-13
Add CMS redirection helper queries (parent + category-of-page)

PrestaEdit:
| Questions         | Answers                                                     |
|-------------------|-------------------------------------------------------------|
| Branch?           | dev                                                          |
| Description?      | Two mono-arg read endpoints used by the front-office redirection routing:<br/>- `GET /cms-page-categories/{cmsPageCategoryId}/parent-for-redirection` exposes `GetCmsPageParentCategoryIdForRedirection` → `{parentId: int}`.<br/>- `GET /cms-pages/{cmsPageId}/category-for-redirection` exposes `GetCmsCategoryIdForRedirection` → `{cmsPageCategoryId: int}`.<br/>Both use the `[_queryResult]` scalar-wrap mapping around the returned `CmsPageCategoryId` VO (mono-arg → normalizer collapses to int). |
| Type?             | new feature                                                  |
| Category?         | CO                                                           |
| BC breaks?        | no                                                           |
| Deprecations?     | no                                                           |
| Fixed ticket?     | Related to PrestaShop/PrestaShop#39630                       |
| How to test?      | Run `composer phpunit-integration`; new `CmsRedirectionQueriesEndpointTest` covers scope protection and both happy paths (root-parent and unknown-page fallback → ROOT id). |

Standalone files under `CmsPageCategory/` and `CmsPage/` folders to avoid touching in-progress PRs #262 / #193 (CmsPageCategory CRUD) and #165 (CmsPage CRUD).

PR #433, opened 2026-08-20
Merge the CMS endpoints into one domain PR

PrestaEdit:
| Questions         | Answers
| Branch?           | dev
| Description?      | Consolidates the five pending CMS PRs into one domain PR, and moves the read-only tests off the ROOT category onto fixtures they create
| BC breaks?        | no
| Deprecations?     | no

## What this PR does

Merges #262, #270, #362, #375 and #380 into one PR, following the mutualisation done on the Product domain in #410.

### Endpoints

| Method | URI | Scope |
|--------|-----|-------|
| GET / POST / PATCH / DELETE | `/cms-page-categories`, `/cms-page-categories/{cmsPageCategoryId}` | `cms_page_category_read` / `_write` |
| PUT | `/cms-page-categories/{cmsPageCategoryId}/toggle-status` | `cms_page_category_write` |
| PUT | `/cms-page-categories/bulk-enable`, `/cms-page-categories/bulk-disable` | `cms_page_category_write` |
| DELETE | `/cms-page-categories/bulk-delete` | `cms_page_category_write` |
| GET / POST / PATCH / DELETE | `/cms-pages`, `/cms-pages/{cmsPageId}` | `cms_page_read` / `_write` |
| PUT | `/cms-pages/{cmsPageId}/toggle-status` | `cms_page_write` |
| PUT | `/cms-pages/bulk-enable`, `/cms-pages/bulk-disable` | `cms_page_write` |
| DELETE | `/cms-pages/bulk-delete` | `cms_page_write` |
| GET | `/cms-pages/{cmsPageId}/category-for-redirection` | `cms_page_read` |

### Why CmsPage and CmsPageCategory are one PR

#362 spans both namespaces on its own: `CmsPageCategoryForRedirection` addresses a `cmsPageId` and lives under `CmsPage`, `CmsPageCategoryParentForRedirection` addresses a `cmsPageCategoryId` and lives under `CmsPageCategory`. Splitting the domain in two would have meant splitting that PR in two as well, for two queries that answer the same question from either side. The namespaces are right — each URI matches its subject — but the PR boundary was not.

### Three of #362's endpoints are dropped

Consolidating them with a create endpoint is what made them testable for the first time, and all three turned out not to work:

- **`GET /cms-page-categories/{id}/localized-names`** — `GetCmsPageCategoryNameForListing` takes no identifier at all. Its handler reads `id_cms_category` off the current request and falls back to the root category, so the endpoint answered `"Home"` whatever id the URI carried. It is a back office listing helper.
- **`GET /cms-page-categories/{id}/breadcrumbs`** — `GetCmsPageCategoriesForBreadcrumb` has an untyped constructor, so the id from the URI reaches `CmsPageCategoryId` as a string and its `is_int()` assertion rejects it: the endpoint answered `500` on every call. Its `Breadcrumb` result is iterator-only, and its items expose a `CmsPageCategoryId` object rather than a scalar.
- **`GET /cms-page-categories/{id}/parent-for-redirections`** — `GET /cms-page-categories/{id}` already returns `parentId`. Exposing it again also collides: the query returns a `CmsPageCategoryId`, which normalizes to the very key the URI identifier already occupies, so the response carried the identifier alone.

The three queries are registered in `EXCLUDED_CQRS_CLASSES` with those reasons, the way the quick access commands are, so the tracking table stops counting them as missing endpoints.

`GET /cms-pages/{id}/category-for-redirections` stays: its identifier is `cmsPageId`, so the `CmsPageCategoryId` the query returns lands on a free key, and it has behaviour of its own — an unknown page falls back to the root category.

### The two toggles return the entity

`toggle-status` answered `204` on both resources while the `GET` right above returned the entity. Both now replay the query of the `GET` — `GetCmsPageForEditing` and `GetCmsPageCategoryForEditing` — so a toggle answers `200` with what the following `GET` reports, which the tests assert.

### The read-only PRs were all testing the fixtures

None of #362, #375 and #380 had a create endpoint of its own, so all three queried the ROOT category (id 1), or a page id chosen to be missing. With #262 and #270 in the same PR, each of them now works on something the suite created:

| Was | Now |
|-----|-----|
| `category-for-redirection` on `/cms-pages/999999/...`, covering only the fallback to ROOT | the nominal case on a page created in a known category — the fallback keeps its own test |

That last one is worth spelling out: **the only assertion #362 could make about `category-for-redirection` was about its error path.** The endpoint's actual behaviour was never exercised.

## How to test

```
POST /cms-page-categories                              -> 201
POST /cms-pages                                        -> 201
GET  /cms-pages/{id}/category-for-redirection          -> 200 {cmsPageId, cmsPageCategoryId}
```

Covered by `CmsPageCategoryEndpointTest` and `CmsPageEndpointTest`.

## Supersedes

- #262
- #270
- #362
- #375
- #380

All five will be closed once the CI is green here.


boo-code:
The first of the three integration failures has a non-obvious cause, worth stating before it gets chased as a test bug.

**`GET /cms-page-categories/{cmsPageCategoryId}/localized-names` cannot read the id from the path.** `GetCmsPageCategoryNameForListing` is an empty query class - no constructor arguments - so the normalizer has nothing to map `{cmsPageCategoryId}` onto. The handler resolves the category itself:

```php
$categoryIdFromRequest = $currentRequest->query->getInt('id_cms_category');
return $categoryIdFromRequest ?: CmsPageCategoryId::ROOT_CMS_PAGE_CATEGORY_ID;
```

`$request->query` is the query string only - a path variable lands in `$request->attributes` and is never seen. So the endpoint always answers for category 1, which is what the run shows:

```
testGetCmsPageCategoryLocalizedName
-'Child localized-name EN'
+'Home'
```

That is also why the three source PRs looked green: they queried ROOT and asserted `assertIsString($result['name'])`, which holds whatever comes back. Moving the test onto a category the suite creates is what exposed it. The test is right; the endpoint is not, and no change to the test will fix it - the core query has to take a `CmsPageCategoryId`, the same kind of dependency #432 documents for PrestaShop/PrestaShop#41742.

While the resource is being looked at: it returns `public string $name` in the context language, so `localized-names` promises a per-language map it does not deliver, and that URI is BC-locked once released.

The second failure is unrelated to the first. `Invalid cms page category id '4'` comes from `CmsPageCategoryId::assertIsIntegerGreaterThanZero` receiving `["string","4"]` - the URI variable reaches `GetCmsPageCategoriesForBreadcrumb::__construct` uncast, so that one is a serializer/type-cast issue rather than a missing argument.

nicosomb (commented):
Two points from the automated pre-review don't hold up:

CmsPageCategory QUERY_MAPPING: EditableCmsPageCategory already exposes getShopIds(), not getShopAssociation() like EditableCmsPage does, so shopIds already matches the DTO property name and needs no mapping entry. No fix needed, though testGetCmsPageCategory doesn't assert shopIds either way, worth adding.

CmsPage CREATE_COMMAND_MAPPING vs UPDATE_COMMAND_MAPPING asymmetry: intentional, not a bug. AddCmsPageCommand's constructor parameters are named indexedForSearch and displayed, matching the API fields directly, while EditCmsPageCommand only exposes setIsIndexedForSearch and setIsDisplayed. Adding the same entries to CREATE_COMMAND_MAPPING would break creation.

src/ApiPlatform/Resources/CmsPageCategory/BulkDeleteCmsPageCategory.php
```diff
@@ -0,0 +1,54 @@
...
+            uriTemplate: '/cms-page-categories/bulk-delete',
+            CQRSCommand: BulkDeleteCmsPageCategoryCommand::class,
+            scopes: [
+                'cms_page_category_write',
+            ],
+            allowEmptyBody: false,
+        ),
+    ],
+    exceptionToStatus: [
+        CmsPageCategoryException::class => Response::HTTP_NOT_FOUND,
```
nicosomb, line 43: CmsPageCategoryException is the base exception. The Cannot*CmsPageCategoryException classes thrown by the handler, and CmsPageCategoryConstraintException thrown by the command on invalid payloads, both extend it directly rather than each other, so mapping only the base class here sends every failure back as 404, including real constraint violations. Same pattern as CmsPage.php would fix it: CmsPageCategoryNotFoundException to 404, CmsPageCategoryException to 422.

src/ApiPlatform/Resources/CmsPageCategory/BulkUpdateStatusCmsPageCategory.php
```diff
@@ -0,0 +1,63 @@
...
+            uriTemplate: '/cms-page-categories/bulk-disable',
+            output: false,
+            CQRSCommand: BulkDisableCmsPageCategoryCommand::class,
+            scopes: [
+                'cms_page_category_write',
+            ],
+        ),
+    ],
+    exceptionToStatus: [
+        CmsPageCategoryException::class => Response::HTTP_NOT_FOUND,
```
nicosomb, line 52: Same issue as BulkDeleteCmsPageCategory: mapping the base CmsPageCategoryException to 404 catches real constraint violations too (e.g. CmsPageCategoryConstraintException thrown by the command on invalid ids). Should map NotFoundException to 404 and the base exception to 422, like CmsPage.php does.

src/ApiPlatform/Resources/CmsPage/CmsPage.php
```diff
@@ -0,0 +1,181 @@
...
+            validationContext: ['groups' => ['Default', 'Create']],
+            CQRSCommand: AddCmsPageCommand::class,
+            CQRSQuery: GetCmsPageForEditing::class,
+            scopes: [
+                'cms_page_write',
+            ],
+            CQRSQueryMapping: self::QUERY_MAPPING,
+            CQRSCommandMapping: self::CREATE_COMMAND_MAPPING,
+        ),
+        new CQRSPartialUpdate(
```
nicosomb, line 63: This operation has no validationContext, unlike CQRSCreate above. DefaultLanguage(groups: ['Update']) on titles and linkRewrites below won't be evaluated on PATCH since only the Default group is validated when none is set. Needs validationContext: ['groups' => ['Default', 'Update']].

src/ApiPlatform/Resources/CmsPageCategory/CmsPageCategory.php
```diff
@@ -0,0 +1,158 @@
...
+            validationContext: ['groups' => ['Default', 'Create']],
+            CQRSCommand: AddCmsPageCategoryCommand::class,
+            CQRSQuery: GetCmsPageCategoryForEditing::class,
+            scopes: [
+                'cms_page_category_write',
+            ],
+            CQRSQueryMapping: self::QUERY_MAPPING,
+            CQRSCommandMapping: self::COMMAND_MAPPING,
+        ),
+        new CQRSPartialUpdate(
```
nicosomb, line 63: Same as CmsPage.php: no validationContext here, so DefaultLanguage(groups: ['Update']) on names and linkRewrites never actually validates on PATCH.

src/ApiPlatform/Resources/CmsPage/CmsPageList.php
```diff
@@ -0,0 +1,58 @@
...
+            ],
+        ),
+    ]
+)]
+class CmsPageList
+{
+    #[ApiProperty(identifier: true)]
+    public int $cmsPageId;
+
+    public string $metaTitle;
```
nicosomb, line 49: CmsPageQueryBuilder also selects active, position and id_cms_category (plus head_seo_title), none of which are exposed here. Worth checking if that's intentional or an oversight.

src/ApiPlatform/Resources/CmsPageCategory/CmsPageCategoryList.php
```diff
@@ -0,0 +1,56 @@
...
+            ],
+        ),
+    ]
+)]
+class CmsPageCategoryList
+{
+    #[ApiProperty(identifier: true)]
+    public int $cmsPageCategoryId;
+
+    public string $name;
```
nicosomb, line 49: CmsPageCategoryQueryBuilder also selects active and position, not exposed here.

PR #435, opened 2026-09-09
Add the extra property definition endpoints

jolelievre:
CRUD and list endpoints for the extra property definitions registry, on top of the core CQRS layer (PrestaShop/PrestaShop#41542).

Depends on the core PR https://github.com/PrestaShop/PrestaShop/pull/42787: the commands must accept the constraint DSL string, and the union-typed `defaultValue` needs the scalar-first type extractor. Until that core PR is merged the three new test classes fail on the `9.2.x` / `develop` CI jobs and are skipped on older cores (`minVersion 9.2.0`).

## Endpoints

| Operation | URI | Scope |
|---|---|---|
| `CQRSGet` | `GET /extra-property-definitions/{extraPropertyDefinitionId}` | `extra_property_definition_read` |
| `CQRSCreate` | `POST /extra-property-definitions` | `extra_property_definition_write` |
| `CQRSPartialUpdate` | `PATCH /extra-property-definitions/{extraPropertyDefinitionId}` | `extra_property_definition_write` |
| `CQRSDelete` | `DELETE /extra-property-definitions/{extraPropertyDefinitionId}` (optional body `{"dropColumn": true}`) | `extra_property_definition_write` |
| `PaginatedList` | `GET /extra-property-definitions` (grid `extra_property_definition`) | `extra_property_definition_read` |
| `CQRSDelete` | `DELETE /extra-property-definitions/bulk-delete` | `extra_property_definition_write` |

Payload notes:

- `constraints` is the extra property constraint DSL string (`NotBlank\nLength(min: 2, max: 64)`), read back in its canonical form; `""` on PATCH removes every constraint.
- `type`, `scope`, `sqlIndex` are the enum string values, documented as OpenAPI enums.
- `defaultValue` is typed like the property: `int|string|bool|DecimalNumber|null` (a float default travels as a JSON number).
- `shopIds` maps the shop association (`null` = fallback, `[]` on PATCH reverts to it).
- Module-owned definitions are readable and listed; PATCH/DELETE on them answer 422, except a `shopIds` change.

## Tests

- `ExtraPropertyDefinitionEndpointTest`: contract of the endpoints (typed defaults, canonical constraints, partial update, list filters, delete keeping vs dropping the column, bulk 207, one error per HTTP mapping, module-owned protection). The CQRS error matrix itself is covered by the core Behat suite.
- `ExtraPropertyValuesEndpointTest`: runtime effect of API-created definitions on the real product endpoints: defaults for a row-less product, JSON decoded, write and read back, constraint violations at `extraProperties._core.<property>`, inline `extra__core_<property>` keys on the grid-backed list and on the CQRS combination list.
- `ExtraPropertyMultiShopEndpointTest`: shop-scoped values follow the shop context (single shop, `allShops` fan-out) and a definition restricted with `shopIds` only exists on its shops, on the product endpoint and in the definition list.

Run locally against a 9.2.x core: `composer run-module-tests` or `_PS_ROOT_DIR_=<core> php vendor/phpunit/phpunit/phpunit -c tests/Integration/phpunit-local.xml --filter ExtraProperty`.

### DIFF
--- a/src/ApiPlatform/Resources/ExtraPropertyDefinition/ExtraPropertyDefinition.php
+++ b/src/ApiPlatform/Resources/ExtraPropertyDefinition/ExtraPropertyDefinition.php
@@ -1,0 +1,289 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\Module\APIResources\ApiPlatform\Resources\ExtraPropertyDefinition;
+
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ApiResource;
+use PrestaShop\Decimal\DecimalNumber;
+use PrestaShop\PrestaShop\Core\Domain\ExtraProperty\Command\AddExtraPropertyDefinitionCommand;
+use PrestaShop\PrestaShop\Core\Domain\ExtraProperty\Command\DeleteExtraPropertyDefinitionCommand;
+use PrestaShop\PrestaShop\Core\Domain\ExtraProperty\Command\UpdateExtraPropertyDefinitionCommand;
+use PrestaShop\PrestaShop\Core\Domain\ExtraProperty\Exception\ExtraPropertyConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\ExtraProperty\Exception\ExtraPropertyDefinitionNotFoundException;
+use PrestaShop\PrestaShop\Core\Domain\ExtraProperty\Exception\ExtraPropertyRegistrationFailureException;
+use PrestaShop\PrestaShop\Core\Domain\ExtraProperty\Exception\ProtectedModuleExtraPropertyDefinitionException;
+use PrestaShop\PrestaShop\Core\Domain\ExtraProperty\Query\GetExtraPropertyDefinitionForEditing;
+use PrestaShop\PrestaShop\Core\ExtraProperty\Exception\InvalidExtraPropertyDefinitionException;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSCreate;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSDelete;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSGet;
+use PrestaShopBundle\ApiPlatform\Metadata\CQRSPartialUpdate;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Validator\Constraints as Assert;
+
+/**
+ * Registry of the "extra property" definitions: the typed fields a module (or the merchant, through
+ * this API) adds to native entities. Definitions created here are core-owned (no moduleName);
+ * module-owned definitions are readable but only their shop association can be modified.
+ *
+ * The extra property VALUES themselves are exposed on each entity's own endpoints (the
+ * "extraProperties" sub-object on items, inline "extra_<module>_<property>" keys on lists), not here.
+ */
+#[ApiResource(
+    operations: [
+        new CQRSGet(
+            uriTemplate: '/extra-property-definitions/{extraPropertyDefinitionId}',
+            requirements: ['extraPropertyDefinitionId' => '\d+'],
+            CQRSQuery: GetExtraPropertyDefinitionForEditing::class,
+            scopes: ['extra_property_definition_read'],
+            CQRSQueryMapping: self::QUERY_MAPPING,
+            extraProperties: self::VERSION_GATE,
+        ),
+        new CQRSCreate(
+            uriTemplate: '/extra-property-definitions',
+            validationContext: ['groups' => ['Default', 'Create']],
+            CQRSCommand: AddExtraPropertyDefinitionCommand::class,
+            CQRSQuery: GetExtraPropertyDefinitionForEditing::class,
+            scopes: ['extra_property_definition_write'],
+            CQRSQueryMapping: self::QUERY_MAPPING,
+            CQRSCommandMapping: self::COMMAND_MAPPING,
+            extraProperties: self::VERSION_GATE,
+        ),
+        new CQRSPartialUpdate(
+            uriTemplate: '/extra-property-definitions/{extraPropertyDefinitionId}',
+            requirements: ['extraPropertyDefinitionId' => '\d+'],
+            read: false,
+            validationContext: ['groups' => ['Default', 'Update']],
+            CQRSCommand: UpdateExtraPropertyDefinitionCommand::class,
+            CQRSQuery: GetExtraPropertyDefinitionForEditing::class,
+            scopes: ['extra_property_definition_write'],
+            CQRSQueryMapping: self::QUERY_MAPPING,
+            CQRSCommandMapping: self::COMMAND_MAPPING,
+            extraProperties: self::VERSION_GATE,
+        ),
+        new CQRSDelete(
+            uriTemplate: '/extra-property-definitions/{extraPropertyDefinitionId}',
+            requirements: ['extraPropertyDefinitionId' => '\d+'],
+            output: false,
+            CQRSCommand: DeleteExtraPropertyDefinitionCommand::class,
+            scopes: ['extra_property_definition_write'],
+            CQRSCommandMapping: self::COMMAND_MAPPING,
+            extraProperties: self::VERSION_GATE,
+            // The body is optional: without it the storage column (and its data) is kept, like the
+            // back office "Delete" action; {"dropColumn": true} also drops the column.
+            openapiContext: [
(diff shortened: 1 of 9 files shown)


src/ApiPlatform/Resources/ExtraPropertyDefinition/ExtraPropertyDefinition.php
```diff
@@ -0,0 +1,289 @@
...
+            CQRSQuery: GetExtraPropertyDefinitionForEditing::class,
+            scopes: ['extra_property_definition_write'],
+            CQRSQueryMapping: self::QUERY_MAPPING,
+            CQRSCommandMapping: self::COMMAND_MAPPING,
+            extraProperties: self::VERSION_GATE,
+        ),
+        new CQRSPartialUpdate(
+            uriTemplate: '/extra-property-definitions/{extraPropertyDefinitionId}',
+            requirements: ['extraPropertyDefinitionId' => '\d+'],
+            read: false,
```
jolelievre: Why do you force this? the CQRSPartialUpdate sets it to false automatically by default
Maybe you are using old examples as the base that should be fixed? Or the module's agent context files need an update?

src/ApiPlatform/Resources/ExtraPropertyDefinition/ExtraPropertyDefinition.php
```diff
@@ -0,0 +1,289 @@
...
+            CQRSQuery: GetExtraPropertyDefinitionForEditing::class,
+            scopes: ['extra_property_definition_write'],
+            CQRSQueryMapping: self::QUERY_MAPPING,
+            CQRSCommandMapping: self::COMMAND_MAPPING,
+            extraProperties: self::VERSION_GATE,
+        ),
+        new CQRSDelete(
+            uriTemplate: '/extra-property-definitions/{extraPropertyDefinitionId}',
+            requirements: ['extraPropertyDefinitionId' => '\d+'],
+            output: false,
```
jolelievre: Same here; this is the default value set automatically already? Another example that needs some updating, maybe?

src/ApiPlatform/Resources/ExtraPropertyDefinition/ExtraPropertyDefinition.php
```diff
@@ -0,0 +1,289 @@
...
+    #[ApiProperty(openapiContext: ['type' => 'array', 'items' => ['type' => 'integer'], 'nullable' => true])]
+    public ?array $shopIds;
+
+    public const QUERY_MAPPING = [
+        // URI parameter → GetExtraPropertyDefinitionForEditing(int $id)
+        '[extraPropertyDefinitionId]' => '[id]',
+        // Query result → API resource
+        '[id]' => '[extraPropertyDefinitionId]',
+        '[fieldType]' => '[type]',
+        '[fieldScope]' => '[scope]',
```
jolelievre: Hmm, this should probably be fixed on the core while we still can
There's no reason the CQRS layer would be the only one to prefix those two properties with `field` It's not present here, and it's not presetn in the core ExtraProperty namespace
So let's try and maintain consistency across all layers while we still can

tests/Integration/ApiPlatform/ExtraPropertyDefinitionEndpointTest.php
```diff
@@ -0,0 +1,425 @@
...
+        $this->assertSame('small', $choice['defaultValue']);
+    }
+
+    /**
+     * @depends testAddExtraPropertyDefinition
+     */
+    public function testGetExtraPropertyDefinition(int $definitionId): int
+    {
+        $definition = $this->getItem(self::ENDPOINT . '/' . $definitionId, [self::READ]);
+        $this->assertSame($definitionId, $definition['extraPropertyDefinitionId']);
```
jolelievre: The whole returned object should be tested in a single assertEquals to make sure we check that the read endpoint respects the whole expected contract

tests/Integration/ApiPlatform/ExtraPropertyDefinitionEndpointTest.php
```diff
@@ -0,0 +1,425 @@
...
+        $patchData = [
+            'labelWording' => 'API string (edited)',
+            'displayFront' => false,
+            'required' => true,
+            // Non-destructive storage change: a larger size.
+            'size' => 128,
+            'constraints' => 'Email',
+            'associatedApis' => ['/products/{productId}'],
+        ];
+        $updated = $this->partialUpdateItem(self::ENDPOINT . '/' . $definitionId, $patchData, [self::WRITE]);
```
jolelievre, line 254: Same here we should use a single assertEquals to test it all

tests/Integration/ApiPlatform/ExtraPropertyDefinitionEndpointTest.php
```diff
@@ -0,0 +1,425 @@
...
+        yield 'create endpoint' => ['POST', self::ENDPOINT];
+        yield 'patch endpoint' => ['PATCH', self::ENDPOINT . '/1'];
+        yield 'delete endpoint' => ['DELETE', self::ENDPOINT . '/1'];
+        yield 'list endpoint' => ['GET', self::ENDPOINT];
+        yield 'bulk delete endpoint' => ['DELETE', self::ENDPOINT . '/bulk-delete'];
+    }
+
+    public function testAddExtraPropertyDefinition(): int
+    {
+        $postData = [
```
jolelievre, line 106: Duplictaing the expected data can be annoying, (esepcially since the whole data should be asserted in all following test functions as well) Maybe we can create a function that returns the generic expected data (it can then be modified with the partial changes if needed)

mattgoud:
Tested locally on a 9.2.x docker core running PrestaShop/PrestaShop#42787, with this branch installed through that PR's composer pin (commit `3b884a3`, this PR's head).

## Test results

| Suite | Result |
|---|---|
| `phpunit -c tests/Integration/phpunit-local.xml --filter ExtraProperty` | 28 tests, 354 assertions, OK |
| Full module suite, same config | 638 tests, 5349 assertions, 0 failure, 2 skipped |

So the whole module suite is green against a core that carries the companion PR, not just the three new classes.

## CI failures

The four failing `Integration tests` jobs are the expected cross-PR dependency and nothing else. On `Integration tests (8.1, 9.2.x)` the 16 failures all live in the new `ExtraPropertyDefinitionEndpointTest`, `ExtraPropertyValuesEndpointTest` and `ExtraPropertyMultiShopEndpointTest`, and the first one is the pre-PR `DecimalNumberNormalizer` turning `"n/a"` into a 500 through `PrestaShop\Decimal\Builder::parseNumber`, which is exactly what PrestaShop/PrestaShop#42787 fixes. No other class regresses.

## Review notes

`tests/PHPStan/phpstan-9.2.x.neon` fills a real gap: the workflow matrix already lists `9.2.x` while the config file did not exist, so that job could not have been passing on `dev` before this PR.

The rest reads consistent with the core side:

* `dropColumn` defaults to `false` on both `DeleteExtraPropertyDefinitionCommand` and `BulkDeleteExtraPropertyDefinitionCommand`, which matches the `default: false` documented in the two `openapiContext` blocks;
* the bulk route `/extra-property-definitions/bulk-delete` cannot collide with the item route thanks to the `\d+` requirement;
* the `#Class .*ExtraProperty.* not found.#` ignore is scoped to the resource directory, so it will not mask an unrelated missing class.

Nothing blocking here. The one point I raised on the core PR that is worth a commit is a Behat cleanup issue, which lives on that side.

mattgoud:
I checked the four "items needing closer review" from the pre-review above against the core. Three do not hold, one does.

| Report item | Verdict | Why |
|---|---|---|
| 1. `DecimalNumber` in a union may violate `ApiResourcePropertyTypeRule` | Not an issue | The four PHPStan jobs (9.0.3, 9.1.5, 9.2.x, develop) are green on this branch, so the rule accepts a union mixing scalars with `DecimalNumber`. |
| 2. No `defaultValues` on `CQRSCreate` | **Worth taking** | Documented in `CONTEXT.md` for cores 9.2 and above, and already used by `Carrier.php`. Six fields are optional on the command but documented as required, with their defaults invisible in the schema. |
| 3. `$required` and `$nullable` missing from the list DTO | Not an issue | `ExtraPropertyDefinitionQueryBuilder` selects exactly eight columns (`id_extra_property_definition`, `entity_name`, `module_name`, `property_name`, `type`, `scope`, `sql_index`, `display_front`) and `ExtraPropertyDefinitionList` exposes exactly those eight. No missing property, no orphan property. |
| 4. `type` and `scope` without a `MAPPING` entry | Correct as is | They come back from the SELECT under those exact names. |

## On item 2, one correction before anyone copies the snippet

The report lists `'nullable' => false`, but `AddExtraPropertyDefinitionCommand` declares `bool $nullable = true`. Since the API injects these values into the payload before validation and command denormalization, a wrong default would be worse than none. Read from the command signature:

```php
extraProperties: self::VERSION_GATE + [
    'defaultValues' => [
        'type' => 'string',
        'scope' => 'common',
        'sqlIndex' => 'none',
        'displayFront' => false,
        'required' => false,
        'nullable' => true,
    ],
],
```

## The two remaining suggestions are reasonable, I would take both

- A `getItem` after each `createItem` in `testTypedDefaultValues`. It is the only way the read path of a typed default gets covered.
- Replacing the hardcoded `assertSame(5, $prefixed['totalItems'])` with a count scoped to the definitions that method's own chain creates, so the assertion survives a reordering of the class.

For context on the failing Integration jobs, see my earlier comment: they are the unmerged PrestaShop/PrestaShop#42787 dependency and nothing else.

src/ApiPlatform/Resources/ExtraPropertyDefinition/ExtraPropertyDefinition.php
```diff
@@ -0,0 +1,283 @@
...
+ */
+#[ApiResource(
+    operations: [
+        new CQRSGet(
+            uriTemplate: '/extra-property-definitions/{extraPropertyDefinitionId}',
+            requirements: ['extraPropertyDefinitionId' => '\d+'],
+            CQRSQuery: GetExtraPropertyDefinitionForEditing::class,
+            scopes: ['extra_property_definition_read'],
+            CQRSQueryMapping: self::QUERY_MAPPING,
+            extraProperties: self::VERSION_GATE,
```
Quetzacoalt91, line 60: image

mattgoud (approved):

Quetzacoalt91 (approved):

ps-jarvis:
PR merged, well done!

